PR #2, opened 2017-08-23
Store node.id as signed

labdsf:
Otherwise comparison to -1 converts it to unsigned

### DIFF
--- a/graph.h
+++ b/graph.h
@@ -29,7 +29,7 @@ struct node {
 	uint64_t hash;
 
 	/* ID for .ninja_deps. -1 if not present in log. */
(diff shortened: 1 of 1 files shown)


michaelforney:
This was actually intentional. I check against -1 when assigning IDs, so maybe another option is just to compare against (size_t)-1 instead?

It also looks like ninja just uses int for ID, so int32_t is probably fine as well. However, if we make this change, there are a couple of places in deps.c that would need s/-1/INT32_MAX/.

What are your thoughts?

michaelforney:
I went with the second option. Thanks!

PR #3, opened 2017-08-23
Fix comparison of signed to unsigned in build()

labdsf:
Switched i to size_t like the rest of the functions
and made signed j for 0..maxjobs iterations.


michaelforney:
Thanks, applied.

PR #11, opened 2018-10-08
Reduce an errx to warnx in parse.c to match default Ninja behavior.

ibara:
Hi --

Some of us in OpenBSD land are seriously considering replacing Ninja with Samurai as our default Ninja tool for building third-party packages that use CMake and other Ninja-compatible build tools. See https://marc.info/?l=openbsd-ports&m=153792369811741&w=2 for the thread on the matter. Note that we have already imported Samurai into our ports tree, with me as package maintainer.

In preparation to make that switch, we noticed that there is diverging default behavior between Ninja and Samurai in the event that multiple rules generate a file/target. Ninja by default will warn about this, but continue, providing a
```
ninja: warning: multiple rules generate xxx. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
```
diagnostic message, where `xxx` is the name of the file/target. While Ninja does provide a facility to promote a warning to an error, the default behavior is to warn but continue.

Samurai, in contrast, provides a diagnostic message that reads
```
samu: multiple rules generate 'xxx'
```
and then errors out, providing a return code of 1.

This PR is simply the diff I provided in that mailing list email above, reducing the errx to a warnx to match Ninja's default behavior. With this, there is only a single piece of software, webkitgtk4, that fails to build with Samurai, and which I will look into.

A small repo that exhibits this multiple rules issue is https://github.com/never-lang/never if you'd like to look into it yourself.

Thanks!

### DIFF
--- a/parse.c
+++ b/parse.c
@@ -108,7 +108,7 @@ parseedge(struct environment *env)
 		canonpath(s);
 		*n = mknode(s);
 		if ((*n)->gen)
-			errx(1, "multiple rules generate '%s'", (*n)->path->s);
+			warnx("multiple rules generate '%s'", (*n)->path->s);
 		(*n)->gen = e;
 	}
 


ibara:
Also if you'd like to dig into the corresponding Ninja code, it's here: https://github.com/ninja-build/ninja/blob/ca041d88f4d610332aa48c801342edfafb622ccb/src/manifest_parser.cc#L350

michaelforney:
Wow, it's great to see projects like OpenBSD take a serious look at samurai!

The main reason why multiple generating actions is an error in samurai is that it was not clear to me what the correct behavior was in this case, and it almost always indicates a bug in the build system (I've yet to see otherwise).

In the case of `never`, the build manifest says `parser.txt` is generated by two commands:

    bison -d --report-file=parser.output --verbose -o front/parser.c front/parser.y

and

    cmake -E copy parser.output parser.txt

It is pretty clear that declaring `parser.txt` as an output of the bison command is the problem here. You'll find that running `ninja parser.txt` will never actually build it. I believe that removing these two lines in `FindBISON.cmake` will actually address the issue:
https://gitlab.kitware.com/cmake/cmake/blob/2b836a01217a8e6f14ba5f0d8b89d660e6c5cfca/Modules/FindBISON.cmake#L172-173

That said, I'm not opposed to changing the default behavior to a warning, since you've discovered a good number of packages which don't build because of the error. However, I'm not completely sure of the implications of changing the `errx` to `warnx`. I think there might be some issues in the build logic if there are multiple actions producing the same file.

I took a closer look at how ninja handles this case. It looks like it actually just skips over it when constructing the duplicate action's outputs, making sure to adjust the implicit outputs accordingly. I believe that this branch should have that same behavior:
https://github.com/michaelforney/samurai/tree/multiple-generating-actions

Could you can run the bulk build with that change to see if it fixes your issues?

Regarding the webkitgtk4 build failure, I was unable to reproduce it, but I bet it is just a missing dependency on `jsmin.py` somewhere. Since samurai doesn't always run commands in the same order as ninja, I've occasionally run into build errors due to insufficiently specified dependencies. Probably here[0] and here[1], since both of those scripts import jsmin.

[0] https://github.com/WebKit/webkit/blob/ae64cd8dae83ab8b8b8f3cbdf15eb397b22b598d/Source/WebCore/WebCoreMacros.cmake#L38
[1] https://github.com/WebKit/webkit/blob/ae64cd8dae83ab8b8b8f3cbdf15eb397b22b598d/Source/WebCore/CMakeLists.txt#L1676

michaelforney:
I reported the `BISON_TARGET` issue to cmake here: https://gitlab.kitware.com/cmake/cmake/issues/18433

ibara:
We'd like to do more with Samurai, but it seems like at least part of it was taken from Ninja, and therefore Apache 2.0 licensed. If you/someone ever got around to identifying and rewriting those bits and licensing them ISC like the rest of Samurai, we could talk about doing a lot more with Samurai on OpenBSD.

Back to this issue...
Sure, the changing of errx to warnx is a heavy and likely incomplete hammer. I can get that branch tested in a bulk build, but it will take a week or so--we're actually building the packages that will go with the OpenBSD 6.4 release right now, and so this will have to wait for those packages to be built and distributed first. But for now at least, I'll try your suggestions with webkitgtk4 and see if that fixes things and report back.

michaelforney:
Regarding the licensing:
None of the code is actually taken from ninja, but a large amount of it was written while using the ninja code as a reference (for instance, to fix this bug in a way that is compatible with ninja I had to look at the ninja source since the behavior is not documented). I believe that makes it a derivative.

I detailed the extent of which ninja functions were studied while writing samurai in an email to Marc Espie, but the gist of it is there is likely too much of it to make rewriting those parts feasible.

ibara:
No worries, all good.

I did some manual testing with your new branch and those ports that had multiple rules, and everything appeared to work OK. It's still probably worth a bulk build but I think your fix is right.

michaelforney:
I ended up adding the `-w` flag anyway to try to get samurai to pass some of meson's tests. I think the behavior with multiple generating actions should match ninja's now.

ibara:
Great! Thanks.

Issue #15, opened 2019-01-18
Consider supporting a SAMUFLAGS environment


Earnestly:
This is a feature that `ninja` itself is not willing to support [[0]](https://github.com/ninja-build/ninja/issues/1370) [[1]](https://github.com/ninja-build/ninja/pull/1399) [[2]](https://github.com/ninja-build/ninja/issues/797).

Much like `MAKEFLAGS` these kinds of environments can greatly ease development and maintainence projects.  Though the exposure of such things introduce complication and should generally be avoided it may have benefits outweighing the disadvantages in this case.

I also appreciate that `samurai` itself is trying to maintain feature pairty with `ninja` so this may be rejected on those grounds alone.

Nevertheless, do you have any thoughts about adding such functionality?

Edit: The title should probably be `SAMUFLAGS` to maintain consistency with the binary name.

michaelforney:
I skimmed those discussions, and I think it's unfortunate that so many are frustrated with the outcome.

Here are my current thoughts:
In general my goal with samurai are to remain as simple as possible with supporting most of the essential features of ninja (so it is a drop-in replacement in most cases). I am hesitant to add any features unique to samurai that increase complexity in any way, and also to avoid conflicts in case they are added to a future ninja version with slightly different semantics. Personally, I would have liked to see [explicit scoping], but I have not implemented it for the above reasons.

However, if this were added as a samurai-specific feature (through `SAMUFLAGS`), there is little chance with this conflicting with any future ninja feature. I believe implementation would also be pretty easy. That said, even after reading the arguments in those threads, I still don't really understand why a wrapper script is unacceptable. I don't want to turn this into a rehash of those discussions, but maybe you could provide a concrete example with the system and tools you are using such that a wrapper script does not work?

Regarding the feature itself, I don't see any issue if it is limited to the flags `-v` and `-j`. So, I think it just comes down to whether deviating from ninja's feature set is worth it.

Let me think on this some more.

[explicit scoping]: https://groups.google.com/forum/#!topic/ninja-build/QsJIJU_5blI/discussion

Earnestly:
Yes I agree, I don't wish to rehash any of the arguments here since they're still continuing elsewhere.  Thanks for considering it, all of your points are reasonable and expected.  Although there's always a balance to consider; providing a very common convenience can result in less (potentially broken) total code in the world by adding a degree of complexity to the source (in this case, `FLAGS` environments).

orbea:
[...]

Issue #12, opened 2018-10-29
Windows support


mikejsavage:
Cool project, the problem I have with ninja is you can't static link with musl, and samurai solves that! I'm hoping to have a build system I can add to the repo and forget about so people can just clone + build and it will always work.

I need Windows support, so if I work on that would you be interested in merging it to the main repo?

I took a quick look and it seems like there's not much that needs changing. File I/O stuff and the job spawning/monitoring.

michaelforney:
Yes, I would be interested in merging Windows support. I already have some WIP changes to the lexer and deps/build log to use only C99 library functions.

As you noted, job spawning/monitoring is really the only thing that needs special OS-specific code. I think the best way to do this is to define a generic interface (maybe in an `os.h` header), and then provide implementations for each OS (e.g. `posix.c` and `windows.c`). I definitely do *not* want Windows support added through ifdefs in `build.c`.

> Cool project, the problem I have with ninja is you can't static link with musl, and samurai solves that!

BTW, what issue do you have statically linking ninja with musl? I've done it in the past without issues (before I wrote samurai).

mikejsavage:
Great, would it better if I wait for those changes to go in then? (or post the diff atm and I can take over)

> BTW, what issue do you have statically linking ninja with musl? I've done it in the past without issues (before I wrote samurai).

When I tried I ran into problems with my C++ headers relying on glibc things

michaelforney:
FYI, Zach van Rijn is providing static ninja binaries for Linux and Windows:
https://groups.google.com/d/msg/ninja-build/tPFgdGOAPo8/syQbQILxEQAJ

RoyiAvital:
Is there any progress on that?
Namely having a Samurai binary for Windows?

michaelforney:
The most recent effort was in https://github.com/michaelforney/samurai/pull/37.

I don't have any machines that run Windows, so it's up to whoever wants to do the work.

PR #21, opened 2019-06-29
Implemented SAMUFLAGS

andreasbaumann:
Options in SAMUFLAGS take lower precedence to the options in the command line.
I tried to avoid code duplication as much as possible and allow using the arg.h macros
thus I'm composing a new argv array from the command line options and the parsed
tokens in the SAMUFLAGS environment variable.

### DIFF
--- a/samu.c
+++ b/samu.c
@@ -79,17 +79,60 @@ warnflag(const char *flag)
 		fatal("unknown warning flag '%s'", flag);
 }
 
+enum {
+	maxArgs = 64
+};
+
+static int argc;
+static char *argv_combined[maxArgs];
+static char **argv = &argv_combined[0];
+
+void
+copycmdlineargs(int cmdline_argc, char *cmdline_argv[])
+{
+	int i = 1;
+	
+	while (argc < maxArgs -1 && i <= cmdline_argc) {
+		argv[argc] = cmdline_argv[i];
+		argc++;
+		i++;
+	}
+	argv[argc] = NULL;
+}
+
+void
+parseenvargs(char *s)
+{
+	char *p;
+		
+	p = strtok(s, " ");
+	while (p && argc < maxArgs - 1) {
+		argv[argc++] = p;
+		p = strtok(NULL, " ");
+	}
+	argv[argc] = NULL;
+}
+
 int
-main(int argc, char *argv[])
+main(int cmdline_argc, char *cmdline_argv[])
 {
 	char *builddir, *manifest = "build.ninja", *end, *arg;
 	const struct tool *tool = NULL;
 	struct node *n;
 	long num;
 	int tries;
+	char *samuflags;
 
-	argv0 = strrchr(argv[0], '/');
-	argv0 = argv0 ? argv0 + 1 : argv[0];
+	argv0 = strrchr(cmdline_argv[0], '/');
+	argv0 = argv0 ? argv0 + 1 : cmdline_argv[0];
+	argv[0] = cmdline_argv[0];
+	argc++; cmdline_argc--;
+	samuflags = getenv( "SAMUFLAGS" );
+	if (samuflags != NULL) {
+		samuflags = strdup(samuflags);
+		parseenvargs(samuflags);
+	}
+	copycmdlineargs(cmdline_argc, cmdline_argv);
 	ARGBEGIN {
 	case '-':
 		arg = EARGF(usage());
--- a/samu.c
+++ b/samu.c
@@ -92,28 +92,20 @@ jobsflag(const char *flag)
 }
 
 static void
-parseenvargs(char *s)
+parseenvargs(char *env)
 {
-	char *copy;
-	char *p;
+	char *arg, *argvbuf[64], **argv = argvbuf;
 	int argc;
-	enum { maxArgs = 64 };
-	char *env_argv[maxArgs];
-	char **argv = &env_argv[0];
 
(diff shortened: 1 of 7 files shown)


michaelforney:
Thanks for working on this!

Could you update the manual to mention this new feature?

As mentioned in https://github.com/michaelforney/samurai/issues/15#issuecomment-455711614, I think this feature should be restricted to options like `-v` and `-j N` that don't affect the output of the build. Allowing target names, `-f buildfile`, or `-C dir` in `SAMUFLAGS` could cause confusion at best, and might be a security issue.

I don't think that you need to go out of your way to re-use the `ARGBEGIN`/`ARGEND` macros for handling `SAMUFLAGS`. My suggestion is to just handle the `-v` and `-j` flags in `parseenvargs` directly rather than constructing an additional argv array. Perhaps you could also add a helper to set `buildopts.maxjobs` to ensure that the argument is handled the same way as if it were passed on the command-line.

samu.c
```diff
@@ -79,17 +79,60 @@ warnflag(const char *flag)
...
-	argv0 = argv0 ? argv0 + 1 : argv[0];
+	char *samuflags;
+
+	argv0 = strrchr(cmdline_argv[0], '/');
+	argv0 = argv0 ? argv0 + 1 : cmdline_argv[0];
+	argv[0] = cmdline_argv[0];
+	argc++; cmdline_argc--;
+	samuflags = getenv( "SAMUFLAGS" );
+	if (samuflags != NULL) {
+		samuflags = strdup(samuflags);
```
michaelforney: Does this string need to be duplicated?

If it does, you need to check the result of `strdup`. Note also that `strdup` is not in C99 (though, I believe it will be added to C2X). I'm trying to limit the use of POSIX calls in case someone wants to pick up work on #12. To solve both of these problems, you could use `xmemdup` from `util.c`:

```
samuflags = xmemdup(samuflags, strlen(samuflags) + 1);
```

samu.c
```diff
@@ -79,17 +79,60 @@ warnflag(const char *flag)
...
-
-	argv0 = strrchr(argv[0], '/');
-	argv0 = argv0 ? argv0 + 1 : argv[0];
+	char *samuflags;
+
+	argv0 = strrchr(cmdline_argv[0], '/');
+	argv0 = argv0 ? argv0 + 1 : cmdline_argv[0];
+	argv[0] = cmdline_argv[0];
+	argc++; cmdline_argc--;
+	samuflags = getenv( "SAMUFLAGS" );
```
michaelforney: Please remove the spaces around `"SAMUFLAGS"` to be consistent with the style used for other function calls.

andreasbaumann: ok, done.

samu.c
andreasbaumann: strtok modifies the input string, hence I have to copy the string.

I made a xstrdup helper function using xmemdup function. It also handles the
case of passing a NULL to strdup, which by standard is undefined.

Currently I'm leaving a permanent leak of the copied environment, not sure, how
strict you are on leak checking with valgrind, otherwise I free it at the end, I think
it will solve itself anyway in the next step, when I split of the ARGV and the parseenv
parsing, so I leave it for now.

Though I believe a strdup also exists in the POSIXish parts of Windows
(https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strdup-wcsdup?view=vs-2019) it's best to avoid this function in this case.

samu.c
```diff
@@ -79,6 +79,59 @@ warnflag(const char *flag)
 		fatal("unknown warning flag '%s'", flag);
 }
 
+void
```
michaelforney: This should be `static void` since it is local to this file.

samu.c
```diff
@@ -79,6 +79,59 @@ warnflag(const char *flag)
...
+	case '-':
+		arg = EARGF(usage());
+		if (strcmp(arg, "verbose") == 0) {
+			buildopts.verbose = true;
+		} else {
+			fatal("invalid long argument in SAMUFLAGS");
+		}
+		break;
+	case 'j':
+		num = strtol(EARGF(usage()), &end, 10);
```
michaelforney: Maybe you could separate `-j` handling into a helper function `jobsflag` to make sure it is handled the same way as if it were passed on the command-line (to prevent them from getting out of sync if one changes).

samu.c
```diff
@@ -79,6 +79,59 @@ warnflag(const char *flag)
...
+	char *env_argv[maxArgs];
+	char **argv = &env_argv[0];
+	char *arg, *end;
+	long num;
+		
+	if (s == NULL) {
+		return;
+	}
+
+	copy = xstrdup(s);
```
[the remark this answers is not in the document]
andreasbaumann: ok. I see. I removed the xstrdup helper.

samu.c
andreasbaumann: yep, sorry, forgot. done.

michaelforney:
I hope you don't mind, but I pushed a few changes addressing some minor concerns I had.

I think it's looking pretty good now. My only concern now is the arbitrary limit for number of arguments in `SAMUFLAGS`. Of course, the limit is way more than enough for any use (since only two flags are accepted), so it's probably fine. We should probably report an error though if the limit is hit, rather than silently ignoring some arguments.

michaelforney:
Squashed and pushed now (after a few more tweaks). Thanks for working on this and putting up with my nit-picking.

PR #27, opened 2019-12-07
A modern software requires a modern build system

Salamandar:
Port to meson.
Now, changes in headers will rebuild (only) the required .o files.
I also too the liberty to move source files into a src directory.


michaelforney:
I'm sorry, but the Makefile will stay. Right now, samurai can be built on any POSIX system. The Makefile is strictly POSIX make, and we target C99 (not GNU C99, as you put in meson.build). Using meson means adding a build time dependency, when the Makefile is already sufficient.

What is your main concern here that porting to meson will solve? Is it just rebuilding when headers change? I usually just use the Makefile for bootstrapping and packaging, and the bundled build.ninja for development, which *does* have header dependency tracking.

The main issue with using meson is bootstrapping. Say someone has a plain POSIX system with which they want to build some software using ninja build files. Now, to install samurai, instead of just running `make`, they have to:
- Install python, a run time dependency of meson.
- Install meson, now a build time dependency of samurai.
- Since meson requires an *existing* ninja implementation to build projects, they can't build samurai directly, so they have to install ninja first.
- Ninja is written in C++, so then they have to install a C++ compiler.
- Now, they run ninja's bootstrap.py script to first build ninja, then build it again with itself.
- Finally, they can use meson to generate samurai's ninja build files, then run ninja to build samurai.

Instead, I just updated the Makefile with header dependencies. There isn't much issue with rebuilding everything when a header changes, since this happens rarely, and a complete build of samurai takes a fraction of a second on most systems (0.5 seconds on my 10 year old laptop).

michaelforney:
Also, I can't help but rant:

Meson has been a constant headache for me. They insist on detecting the ninja implementation to use, so I had to make changes to meson (adding to its hard-coded list of ninja command names, and other tweaks) to get it to work with samurai, even though it is none of meson's business which ninja implementation the user plans to run to execute the build, and it would work fine if they didn't do any detection at all.

Additionally, they insist on a hard-coded list of compilers that you can use. So, changing the build system to meson means that I can no longer build samurai with my C compiler, cproc.

I also usually end up hacking up projects meson.build files to get them to build, since my system doesn't have pkg-config files, and meson offers no way to override the detection of packages by specifying the libraries and compile flags as you can with cmake and autotools.

Salamandar:
> Additionally, they insist on a hard-coded list of compilers that you can use. So, changing the build system to meson means that I can no longer build samurai with my C compiler, cproc.
This changed, you can specify compiler and Ninja binary (https://github.com/mesonbuild/meson/pull/5149).

PR #28, opened 2019-12-11
Makefile: Allow overriding PREFIX and MANDIR by build-systems.

ng-0:
improve building in pkgsrc with less SUBST_SED usage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .POSIX:
 
-PREFIX=/usr/local
-MANDIR=$(PREFIX)/share/man
+PREFIX?=/usr/local
+MANDIR?=$(PREFIX)/share/man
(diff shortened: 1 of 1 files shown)


michaelforney:
Not sure why sed would be necessary. Can't you do `make PREFIX=/my/prefix MANDIR=/my/mandir install`?

I avoided `?=` for now, since it is not in the current standard. Though it is accepted for the next version of POSIX (http://austingroupbugs.net/view.php?id=330) so I'm not opposed to this change if it actually fixes a problem.

ng-0:
Oh, I didn't know it was not standard.
I think I can try and provide this as make arguments, I will test this and tell you if it works out this way.

ng-0:
Okay, it works out without this patch.

Thanks!

PR #34, opened 2020-01-26
add support for NINJA_STATUS variable

E5ten:


build.c
```diff
@@ -199,6 +202,46 @@ buildadd(struct node *n)
 	e->flags &= ~FLAG_CYCLE;
 }
 
+static void
+statusfmt(char *fmt)
+{
+	struct timespec ctime;
+
+	for (; *fmt; ++fmt)
```
michaelforney: The brace style I prefer is to use braces if the body is composed of multiple lines (even if they are just a single statement), or if any of the other branches of the if-else chain needed braces.

build.c
```diff
@@ -199,6 +202,46 @@ buildadd(struct node *n)
 	e->flags &= ~FLAG_CYCLE;
 }
 
+static void
+statusfmt(char *fmt)
```
michaelforney: Let's call this `printstatus`.

build.c
```diff
@@ -238,7 +281,8 @@ jobstart(struct job *j, struct edge *e)
 		description = buildopts.verbose ? NULL : edgevar(e, "description", true);
 		if (!description || description->n == 0)
 			description = j->cmd;
-		printf("[%zu/%zu] %s\n", nstarted, ntotal, description->s);
+		statusfmt(ninja_status);
+		printf("%s\n", description->s);
```
michaelforney: `puts(description->s)` is a bit more direct, I think I prefer that.

build.c
```diff
@@ -29,8 +30,10 @@ struct job {
 
 struct buildoptions buildopts = {.maxfail = 1};
 static struct edge *work;
-static size_t nstarted, ntotal;
+static size_t nstarted, nfinished, ntotal;
 static bool consoleused;
+static char *ninja_status;
+static struct timespec stime;
```
michaelforney: Let's call this `starttime`.

build.c
```diff
@@ -199,6 +202,46 @@ buildadd(struct node *n)
...
+				printf("%zu", nstarted - nfinished);
+				break;
+			case 'u':
+				printf("%zu", ntotal - nstarted);
+				break;
+			case 'p':
+				printf("%3lu%%", 100 * nfinished / ntotal);
+				break;
+			case 'e':
+				clock_gettime(CLOCK_REALTIME, &ctime);
```
michaelforney: I think `CLOCK_MONOTONIC` is the right clock to use here. Then it won't be affected by a system clock change during the job (unlikely, but still).

E5ten:
for the default NINJA_STATUS I went with "[%f/%t] " like ninja, but if you want I can change that to "[%s/%t] " to be in line with what samurai does now, it'll also

build.c
```diff
@@ -199,6 +202,46 @@ buildadd(struct node *n)
...
+				printf("%.3f",
+					((double)ctime.tv_sec + (double)ctime.tv_nsec * 0.000000001) -
```
michaelforney: I don't think these casts are necessary. The `0.000000001` literal has type `double`, and usual arithmetic conversions will convert the integers to floating point for the binary operators.

michaelforney:
Also, could you add a short description of `NINJA_STATUS` to the man page?

michaelforney:
Pushed with a few more tweaks. Thanks again!

PR #35, opened 2020-01-28
print output on the same line if output is a terminal

E5ten:
can be manually enabled or disabled at build time by defining
HAVE_TIOCGWINSZ in CFLAGS, otherwise automatically disabled on
operating systems known to have the TIOCGWINSZ ioctl


build.c
```diff
@@ -201,46 +209,52 @@ buildadd(struct node *n)
...
 }
 
-static void
+/* returns length of printed status */
+static int
 printstatus(void)
 {
 	const char *fmt;
 	struct timespec endtime;
+	int len = 0;
```
michaelforney: Even though printf returns `int`, I think it would be better to store the total length in `size_t` here.

I'm also wondering about the case where printf fails and returns -1 (this could happen if, for example, samurai output was redirected to a file on a disk that runs out of space).

build.c
```diff
@@ -495,6 +521,15 @@ jobwork(struct job *j)
 	return false;
 }
 
+#if HAVE_TIOCGWINSZ
+static void
+sig_handler(int sig)
```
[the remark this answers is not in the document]
E5ten: Without checking if sig == SIGWINCH there's a -Wunused-parameter warning and I didn't want to add a warning but I also didn't want to add -Wno-unused-parameter to the compilation flags, any other ideas for how to avoid it?

build.c
E5ten: I'll think about how to deal with the -1 case, but in terms of storing it as a size_t, the parameter for * in a printf format string is supposed to be an int anyway so I'd end up having to cast len to int anyway, wouldn't that remove any benefit of storing it in a size_t?

build.h
```diff
@@ -1,3 +1,12 @@
+#ifndef HAVE_TIOCGWINSZ
+#	if defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) || \
+	defined(__OpenBSD__) || defined(__NetBSD__) || defined(__DragonFly__) || defined(__sun)
+#		define HAVE_TIOCGWINSZ 1
+#	else
+#		define HAVE_TIOCGWINSZ 0
+#	endif
+#endif
```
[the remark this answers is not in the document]
E5ten: Nope, I'll move it to build.c.

build.c
```diff
@@ -282,8 +296,20 @@ jobstart(struct job *j, struct edge *e)
 		description = buildopts.verbose ? NULL : edgevar(e, "description", true);
 		if (!description || description->n == 0)
 			description = j->cmd;
-		printstatus();
+		if (!dumb)
+			putchar('\r');
+		len = printstatus();
+#if HAVE_TIOCGWINSZ
+		if (!dumb && len + description->n > ws.ws_col)
+			printf("%.*s...%s\033[K", (ws.ws_col - 3) / 2 - len, description->s,
```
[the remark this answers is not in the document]
E5ten: I think what I just added deals with this properly.

build.c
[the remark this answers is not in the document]
E5ten: I pushed a set of changes to deal with your suggestions, currently has the -Wunused-parameter I mentioned, so that'll need to be dealt with.

build.c
```diff
@@ -495,6 +521,15 @@ jobwork(struct job *j)
 	return false;
 }
 
+#if HAVE_TIOCGWINSZ
+static void
+sig_handler(int sig)
+{
+	if (sig == SIGWINCH)
+		ioctl(1, TIOCGWINSZ, &ws);
```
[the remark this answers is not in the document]
E5ten: I went with the first option and it seems to work.

build.c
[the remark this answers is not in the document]
michaelforney: I think a common way to deal with this warning is

```c
(void)sig;
```

build.c
michaelforney: Ah, yes, that's a good point.

E5ten: I added a new variable to assign the return value of printf to, and if it's above 0 I add it to len, that should remove possible problems from printf returning -1.

build.c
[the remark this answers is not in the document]
E5ten: Thanks!

build.c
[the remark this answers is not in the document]
E5ten: If you don't think it's a problem to go with the third method, I think that might be preferable just because of the complexity in the code that watching for SIGWINCH has added compared to the one line it would be to check the width for each status printing?

E5ten:
I think I've resolved the issue I tried to solve the wrong way in #44.

build.c
```diff
@@ -555,8 +603,10 @@ build(void)
 		}
 		if (numjobs == 0)
 			break;
-		if (poll(fds, jobslen, -1) < 0)
-			fatal("poll:");
+		while (poll(fds, jobslen, -1) < 0) {
+			if (errno != EINTR)
+				fatal("poll:");
+		}
```
michaelforney, line 609: Does blocking `SIGWINCH` prevent the need for checking for `EINTR` here?

build.c
```diff
@@ -445,8 +483,12 @@ jobdone(struct job *j)
 		j->failed = true;
 	}
 	close(j->fd);
-	if (j->buf.len && (!consoleused || j->failed))
+	if (j->buf.len && (!consoleused || j->failed)) {
+		if (!dumb)
+			putchar('\n');
 		fwrite(j->buf.data, 1, j->buf.len, stdout);
+		newline = true;
```
michaelforney, line 490: Might be good to use `newline = j->buf.data[j->buf.len - 1] == '\n'` to see if the output actually ended in a newline.

build.c
```diff
@@ -290,9 +306,31 @@ jobstart(struct job *j, struct edge *e)
 		description = buildopts.verbose ? NULL : edgevar(e, "description", true);
 		if (!description || description->n == 0)
 			description = j->cmd;
-		formatstatus(status, sizeof(status));
+		if (!dumb)
+			putchar('\r');
+		len = formatstatus(status, sizeof(status));
```
michaelforney, line 311: The return value here is actually the length needed for the full status (similar to how `snprintf`) works. In the highly unlikely case that it is longer than 256 bytes (`len >= sizeof(status)`), you should set `len = sizeof(status) - 1`.

build.c
E5ten: Actually, I think that since I removed the watching for SIGWINCH to rerun ioctl, and instead have it run before each status print, this change shouldn't necessary even without blocking SIGWINCH.

build.c
```diff
@@ -290,9 +306,31 @@ jobstart(struct job *j, struct edge *e)
...
+		ioctl(1, TIOCGWINSZ, &ws);
+		if (!dumb) {
+			if (len + description->n > ws.ws_col) {
+				switch (ws.ws_col) {
+					case 3: putchar('.');  /* fallthrough */
+					case 2: putchar('.');  /* fallthrough */
+					case 1: putchar('.');  /* fallthrough */
+					case 0: break;
+					default: printf("%.*s...%s",
+							(ws.ws_col - 3) / 2 - len, description->s,
```
michaelforney, line 323: Does ninja do anything related to multi-byte character handling here, or does it just assume one byte == one column? If ninja doesn't do anything fancy, than I guess that's okay here, too.

E5ten: As far as I can tell, ninja uses C++'s string::size for this, which apparently returns a byte-count, not taking multi-byte characters into account, so I think it's fine.

E5ten:
> Do you think you could come up with a small abstraction for the status printing? Then the smart status printer could go into a separate file, and we avoid all the ifdefs.
> 
> Maybe something like `statusinit()`, `statusprint(const char *status, const char *description)`, `statusoutput(const char *out, size_t len)`, `statusdone()`?
> 
> Actually, now that I think about it, it might be simpler to just have a single function `size_t termwidth(void)`. term-dumb.c could just return 0, and term-ansi.c could use the ioctl.

What would I do to have the build choose which of term-{dumb,ansi}.c to use?

michaelforney:
> What would I do to have the build choose which of term-{dumb,ansi}.c to use?

I'd just do

```make
TERM=dumb
OBJ=\
	...
	term-$(TERM).o
```

Operating systems that want to enable smart status printing can just build with `make TERM=ansi`.

E5ten:
And should I add a header, or forward declare any functions from term-{dumb,ansi}.c in build.c?

michaelforney:
Yeah, a `term.h` is probably appropriate.

E5ten:
This will probably need at least some cleanup but I've moved the ifdefed stuff into separate files.

term-ansi.c
```diff
@@ -0,0 +1,42 @@
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/ioctl.h>
+#include <unistd.h>
+
+static struct winsize ws;
```
michaelforney, line 8: No need to make this a global variable, right?

build.c
```diff
@@ -259,9 +261,12 @@ jobstart(struct job *j, struct edge *e)
 	struct node *n;
 	struct string *rspfile, *content, *description;
 	int fd[2];
+	unsigned int len;
```
michaelforney, line 264: This should be `size_t`.

term-ansi.c
```diff
@@ -0,0 +1,42 @@
...
+						desc + desclen - (ws.ws_col - 3) / 2);
+			}
+		} else {
+			fputs(desc, stdout);
+		}
+		printf("\033[K");
+		fflush(stdout);
+	} else {
+		puts(desc);
+	}
```
michaelforney, line 41: I'd make this return early to decrease the indentation level:

```c
if (dumb) {
	puts(desc);
	return;
}
```

build.c
```diff
@@ -512,6 +525,8 @@ build(void)
 	size_t i, next = 0, jobslen = 0, numjobs = 0, numfail = 0;
 	struct edge *e;
 
+	dumb = isdumb(buildopts.verbose);
```
michaelforney, line 528: Rather than pass `buildopts.verbose`, I'd just use `dumb = buildopts.verbose || isdumb()` here.

build.c
E5ten: But since it's use is for "%.*s" with printf I'd need to cast it to int anyway, should I change it to size_t and just add a cast anyway?

michaelforney: Yes, I think so. That is an implementation detail of `printdesc`, and should not leak into its interface.

term-dumb.c
```diff
@@ -0,0 +1,19 @@
...
+#include <stdlib.h>
+
+int
+isdumb(void)
+{
+	return 1;
+}
+
+void
+printdesc(bool dumb, int len, size_t desclen, char *desc)
```
michaelforney, line 12: Maybe just pass a `struct string *` here for the description?

build.c
michaelforney: I'd also call it `statuslen` to make it clear what it's a length of.

E5ten:
Is there anything still blocking this?

bonzini:
Honestly I think printing all output on the same line is a bad idea and (the lack of) this is the main reason why I use samu instead of ninja. Since ninja has no "whatif" tool which is the opposite of "commands", that is to print what _would_ be built if a file was touched, the only way to do so is to touch the source file and do a `ninja -n`. But `ninja -n` effectively only prints the last line, so you have to do `ninja -n | cat`... or just use samu instead.

E5ten:
This is a build-time option that defaults to being disabled, so I don't think that's really a problem

PR #37, opened 2020-02-01
DRAFT: Add minimal reproc proof-of-concept integration.

DaanDeMeyer:
I think I've gotten [reproc](https://github.com/DaanDeMeyer/reproc) to the point where it can be used to handle the cross-platform job control of samurai.

I've written a small proof of concept with the minimum amount of changes necessary to replace the current usage of `posix_spawn` with reproc instead. As a test, I'm currently busy compiling LLVM with this PR.

To test these changes:

```
git checkout https://github.com/DaanDeMeyer/samurai.git --branch reproc
meson build
ninja -C build
sudo ln -sf <path-to-samu-in-build-dir> /usr/bin/ninja 
```

As to why I've made this pull request, I'm interested in adding Windows support for Samurai. I'm making this draft pull request mainly to ask if you would be interested in adding reproc as a dependency in order to get Windows support working.

### DIFF
--- a/build.c
+++ b/build.c
@@ -17,13 +17,14 @@
 #include "graph.h"
 #include "log.h"
 #include "util.h"
+#include <reproc/reproc.h>
 
 struct job {
 	struct string *cmd;
 	struct edge *edge;
 	struct buffer buf;
 	size_t next;
-	pid_t pid;
+	reproc_t *process;
 	int fd;
 	bool failed;
 };
@@ -246,13 +247,12 @@ printstatus(void)
 static int
 jobstart(struct job *j, struct edge *e)
 {
-	extern char **environ;
 	size_t i;
 	struct node *n;
 	struct string *rspfile, *content, *description;
 	int fd[2];
-	posix_spawn_file_actions_t actions;
-	char *argv[] = {"/bin/sh", "-c", NULL, NULL};
+	reproc_options options = { 0 };
+	const char *argv[] = {"/bin/sh", "-c", NULL, NULL};
 
 	++nstarted;
 	for (i = 0; i < e->nout; ++i) {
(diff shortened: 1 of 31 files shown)


build.c
```diff
@@ -246,13 +247,12 @@ printstatus(void)
...
 {
-	extern char **environ;
 	size_t i;
 	struct node *n;
 	struct string *rspfile, *content, *description;
 	int fd[2];
-	posix_spawn_file_actions_t actions;
-	char *argv[] = {"/bin/sh", "-c", NULL, NULL};
+	reproc_options options = { 0 };
+	const char *argv[] = {"/bin/sh", "-c", j->cmd->s, NULL};
```
michaelforney: `j->cmd` doesn't get set until later on in this file, so I'm wondering how this is working.

DaanDeMeyer: I never removed the correct assignment later in the file which luckily avoids the issue.

DaanDeMeyer:
Any opinions on how to handle the build system on Windows? I used meson because I wanted to make sure reproc is useable as a meson subproject but CMake is a perfect possibility as well. Make is pretty hard to deal with on Windows when we need to pull in and build a dependency so if we want to stick to make it might be better to copy the necessary sources from reproc into the tree and write the necessary instructions to build reproc with make. I've made sure reproc produces the most trivial command lines possible so integrating it into Make shouldn' t be pretty trivial. However, if we copy reproc's sources into samurai's repo, the dependency issue dissapears and it might be possible to use reproc for both POSIX and Windows. I can also maintain a Makefile in reproc's repository so reproc can be pulled in as a git submodule and included in the main Makefile.

DaanDeMeyer:
I pushed a revision that integrates reproc in the existing Makefile build and copies the relevant sources from reproc into the repository. With this approach, Windows would have its own NMake based Makefile that would compile reproc's Windows implementation instead. Instead of copying reproc's sources into the repository, we could include reproc as a git submodule as well (which would simply require adding `--recursive` when cloning the samurai repository on Windows).

michaelforney:
> Make is pretty hard to deal with on Windows when we need to pull in and build a dependency so if we want to stick to make it might be better to copy the necessary sources from reproc into the tree and write the necessary instructions to build reproc with make.

I'm not too familiar with Windows development. Can you explain what the issues are? Is setting LDFLAGS to point to the library not sufficient?

> However, if we copy reproc's sources into samurai's repo, the dependency issue dissapears and it might be possible to use reproc for both POSIX and Windows. I can also maintain a Makefile in reproc's repository so reproc can be pulled in as a git submodule and included in the main Makefile.

I don't think I want reproc's sources bundled into the samurai repo, but a submodule might be okay. I'm also fine with a separate build system for Windows alongside the POSIX makefile, if that's what's necessary.

I pushed a branch https://github.com/michaelforney/samurai/tree/proc that adds a small abstraction over starting processes. I'm thinking that reproc could be supported by adding a proc-reproc.c, which could be enabled by make PROC=reproc. Of course, the proc.h interface might need tweaking, but hopefully the idea makes sense.

DaanDeMeyer:
I lost track of this. I've done some work on making the non-process OS bits of samurai cross platform. However, I've discovered a pretty nasty reproc issue on Windows that sometimes results in truncated output from child processes if `fclose(stdout)` is not called directly by a child process before exiting. Since we can't control the process that's running as a child process under samurai, we'd be completely dependent on the program we're running to do that for us which is not really acceptable. I've filed a [bug report](https://developercommunity.visualstudio.com/content/problem/977949/child-process-redirect-stdio-to-sockets-stdio-not.html) but it remains to be seen whether it will actually be fixed. If it isn't, using reproc for Windows support becomes a lot more problematic. We'd have to use the ninja subprocess approach but that would require more substantial changes to both reproc and samurai since we'd be forced to adapt to the completion based nonblocking model of Windows which is completely different from the readiness based model of POSIX.

I think it's best to wait until I get a response on that bug report since it influences the best approach to take quite a bit.

michaelforney:
Okay, thanks for the update.

Please let me know if you end up solving that issue. I'm still interested in Windows support via reproc if that ends up being possible.

DaanDeMeyer:
Alright, unfortunately, it doesn't seem that Microsoft is interested in fixing this particular issue. Luckily, I've found a workaround by keeping extra handles to the child process output pipes alive in the parent process and closing those when the child process exits. In the parent process, I can guarantee that the pipes are closed correctly and no data is lost. The minor disadvantage is that you only get a `POLLHUP` for the output pipes on Windows when the child process exits (even if the child process closes them earlier). For most programs, I don't think this is a big problem compared to the data loss we had previously. 

I'll see when I can find more time to work on this and I'll re-open the PR once I have some more updates.

PR #42, opened 2020-02-25
tool: Implement targets

### LINKED ISSUE #39: Support for samu -t targets
When using `meson` on a big project, I often like to know the available targets, but the feature seems to not be present in `samurai`.

It's not required by any build and not blocking my work, so the priority is pretty low.

[ninja implementation][1]

[1]: https://github.com/ninja-build/ninja/blob/d986e4db5630cf1c5547e69b5556f006f7d3444a/src/ninja.cc#L418

Bl4ckb0ne:
Implements the tool `targets`, fixes #39.

So far `rule` and `all` are implemented, I'm experiencing trouble with `depth`, as the field `nuse` of every node of every edge is always 0. Plus `nodes[i].gen->rule->name` (tool.c:255) seems to not be initialized yet.

@michaelforney any pointers for my `nuse` issue?

### DIFF
--- a/graph.c
+++ b/graph.c
@@ -226,28 +226,3 @@ edgeadddeps(struct edge *e, struct node **deps, size_t ndeps)
 	e->inorderidx += ndeps;
 	e->nin += ndeps;
 }
-
-size_t
-rootnodes(struct node **nodes)
-{
-	size_t size = 0;
-	struct edge *e = NULL;
-
-	for (e = alledges; e; e = e->allnext) {
-		if (e->nout == 0)
-			continue;
-
-		for (size_t n = 0; n < e->nout; ++n) {
-			printf("rule is %s\n", e->rule->name);
-			printf("nuse is %zu\n", e->out[n]->nuse);
-			if (e->out[n]->nuse == 0) {
-				*nodes = xreallocarray(*nodes, size + 1, sizeof(nodes[0]));
-				nodes[size] = e->out[n];
-				size += 1;
-			}
-		}
-
-	}
-
-	return size;
-}
--- a/graph.c
+++ b/graph.c
@@ -226,3 +226,28 @@ edgeadddeps(struct edge *e, struct node **deps, size_t ndeps)
 	e->inorderidx += ndeps;
 	e->nin += ndeps;
 }
+
+size_t
+rootnodes(struct node **nodes)
+{
+	size_t size = 0;
+	struct edge *e = NULL;
+
+	for (e = alledges; e; e = e->allnext) {
+		if (e->nout == 0)
+			continue;
+
+		for (size_t n = 0; n < e->nout; ++n) {
+			printf("rule is %s\n", e->rule->name);
+			printf("nuse is %zu\n", e->out[n]->nuse);
+			if (e->out[n]->nuse == 0) {
+				*nodes = xreallocarray(*nodes, size + 1, sizeof(nodes[0]));
+				nodes[size] = e->out[n];
+				size += 1;
+			}
+		}
+
+	}
+
+	return size;
+}
--- a/samu.1
+++ b/samu.1
@@ -24,8 +24,13 @@
 .Op Fl C Ar dir
 .Op Fl f Ar buildfile
 .Fl t Cm compdb
-.Op Fl x
 .Op Ar rule...
+.Nm
+.Op Fl C Ar dir
+.Op Fl f Ar buildfile
+.Fl t Cm targets
+.Op Ar mode
+.Op Ar argument
 .Sh DESCRIPTION
 .Nm
 loads a build manifest from
@@ -60,6 +65,17 @@ If the
 tool is used, a compilation database
 .Lk ( https://clang.llvm.org/docs/JSONCompilationDatabase.html )
 is printed instead.
+.Pp
+If the
+.Cm targets
+tool is used, a list of targets will be displayed, either by rule or by depth.
+If a name is given as an argument to the rule mode, a list of targets using the
+given rule will be displayed. If no name is given, a list of source files is
+displayed. If a value is given as an argument to the depth mode, a list of
+targets will be displayed in a depth-first way, starting by the root targets
+(the targets with no outputs). If the depth is 0, all targets will be displayed.
+If no arguments is provided to the depth mode, a depth of 1 will be assumed. If
+no mode is specified, a list of all targets is displayed.
 .Sh OPTIONS
 .Bl -tag -width Ds
 .It Fl C
--- a/samu.1
+++ b/samu.1
@@ -24,6 +24,7 @@
 .Op Fl C Ar dir
 .Op Fl f Ar buildfile
 .Fl t Cm compdb
+.Op Fl x
 .Op Ar rule...
 .Nm
 .Op Fl C Ar dir
--- a/samu.1
+++ b/samu.1
@@ -71,12 +71,14 @@ If the
 .Cm targets
 tool is used, a list of targets will be displayed, either by rule or by depth.
 If a name is given as an argument to the rule mode, a list of targets using the
-given rule will be displayed. If no name is given, a list of source files is
-displayed. If a value is given as an argument to the depth mode, a list of
-targets will be displayed in a depth-first way, starting by the root targets
-(the targets with no outputs). If the depth is 0, all targets will be displayed.
-If no arguments is provided to the depth mode, a depth of 1 will be assumed. If
-no mode is specified, a list of all targets is displayed.
+given rule will be displayed.
+If no name is given, a list of source files is displayed.
+If a value is given as an argument to the depth mode, a list of targets will be
+displayed in a depth-first way, starting by the root targets (the targets with
+no outputs).
+If the depth is 0, all targets will be displayed.
+If no arguments is provided to the depth mode, a depth of 1 will be assumed.
+If no mode is specified, a list of all targets is displayed.
 .Sh OPTIONS
 .Bl -tag -width Ds
 .It Fl C
--- a/tool.c
+++ b/tool.c
@@ -243,20 +243,21 @@ targets_rule(int argc, char *argv[])
 }
 
 static void
-targets_depth(struct node **nodes, size_t nnodes, size_t depth, size_t indent)
+targets_depth(struct node *n, size_t depth, size_t indent)
 {
-	size_t i, j;
-	for (i = 0; i < nnodes; ++i) {
-		if (nodes[i]->gen == NULL)
-			continue;
+	struct edge *e = n->gen;
+	size_t i;
 
-		for (j = 0; j < indent; ++j)
-			printf("  ");
-
-		printf("%s: %s\n", nodes[i]->path->s, nodes[i]->gen->rule->name);
-
-		if (depth > 1 || depth <= 0)
-			targets_depth(nodes[i]->gen->in, nodes[i]->gen->nin, depth - 1, indent + 1);
+	for (i = 0; i < indent; ++i)
+		printf("  ");
+	if (e) {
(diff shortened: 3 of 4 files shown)


tool.c
```diff
@@ -221,9 +222,83 @@ compdb(int argc, char *argv[])
...
+static void
+targets_rule(int argc, char *argv[])
+{
+	struct edge *e = NULL;
+	for (e = alledges; e; e = e->allnext) {
+		if (argc == 2) {
+			if (e->nin == 0)
+				continue;
+
+			/* TODO: what does `gen` means in a node? */
```
michaelforney: `gen` is the edge that generates the node (the build action which produces the node as an output).

graph.c
```diff
@@ -225,3 +225,28 @@ edgeadddeps(struct edge *e, struct node **deps, size_t ndeps)
 	e->inorderidx += ndeps;
 	e->nin += ndeps;
 }
+
+size_t
+rootnodes(struct node **nodes)
```
michaelforney: Rather than adding this `rootnodes` function, what do you think about `targets_depth` looping through nodes with `nuse == 0` directly, to avoid the intermediate array?

Bl4ckb0ne: I thought it would be nice to have the function available for future uses, but if it's the only use I will merge it into `targets_depth`.

michaelforney:
The `nuse` issue should be solved as of c02b5bba6f39fbe6dc65dd1ac12176333ae6a10b. Let me know if you run into any other problems.

Bl4ckb0ne:
Thanks! I'll rebase my branch right away.

graph.c
Bl4ckb0ne: I merged the `targets` function.

Bl4ckb0ne:
Alright @michaelforney, I think I'm ready for review. The spec about the tool `targets` is [here](https://github.com/ninja-build/ninja/blob/master/doc/manual.asciidoc)

> targets | output a list of targets either by rule or by depth.  If used like ninja -t targets rule name it prints the list of targets using the given rule to be built.  If no rule is given, it prints the source files (the leaves of the graph).  If used like ninja -t targets depth digit it prints the list of targets in a depth-first manner starting by the root targets (the ones with no outputs). Indentation is used to mark dependencies. If the depth is zero it prints all targets. If no arguments are provided ninja -t targets depth 1 is assumed. In this mode targets may be listed several times. If used like this ninja -t targets all it prints all the targets available without indentation and it is faster than the depth mode.
> -- | --

I made some tests by hand against `ninja` on the `samurai` repo, here's the [log](https://paste.sr.ht/%7Ebl4ckb0ne/e39583b2dd47e92c65e41dfef4133f9ffd97e28b)

tool.c
```diff
@@ -221,9 +222,105 @@ compdb(int argc, char *argv[])
 	return 0;
 }
 
+static void
+targets_rule(int argc, char *argv[])
+{
+	struct edge *e = NULL;
```
michaelforney: No need to initialize this to `NULL`, also please add a blank line separating the initial declarations.

tool.c
```diff
@@ -221,9 +222,105 @@ compdb(int argc, char *argv[])
...
+	for (e = alledges; e; e = e->allnext) {
+		if (argc == 2) {
+			if (e->nin == 0)
+				continue;
+
+			if (!e->in[0]->gen)
+				printf("%s\n", e->in[0]->path->s);
+		}
+		else if (argc == 3) {
```
michaelforney: This should be merged with the previous line with the closing `}`.

tool.c
```diff
@@ -221,9 +222,105 @@ compdb(int argc, char *argv[])
...
+			if (e->nout == 0)
+				continue;
+
+			if (strcmp(e->rule->name, argv[2]) == 0)
+				printf("%s\n", e->out[0]->path->s);
+		}
+	}
```
michaelforney: I think you need to iterate through all inputs/outputs of the edge, not just the first one.

tool.c
```diff
@@ -221,9 +222,105 @@ compdb(int argc, char *argv[])
...
+
+static void
+targets_all(void)
+{
+	struct edge *e = NULL;
+	for (e = alledges; e; e = e->allnext) {
+		if (e->nout == 0)
+			continue;
+
+		printf("%s: %s\n", e->out[0]->path->s, e->rule->name);
```
michaelforney: This also should iterate through all outputs of the edge.

tool.c
```diff
@@ -221,9 +222,105 @@ compdb(int argc, char *argv[])
...
+{
+	size_t depth = 1;
+	if (argc >= 2) {
+		const char *mode = argv[1];
+		if (strcmp(mode, "rule") == 0) {
+			targets_rule(argc, argv);
+			return 0;
+		} else if (strcmp(mode, "depth") == 0) {
+			if (argc > 1)
+				sscanf(argv[2], "%zu", &depth);
```
michaelforney: Use `strtol` here so that we can detect when we are passed an invalid depth. See the argument handling in samu.c for examples.

tool.c
```diff
@@ -221,9 +222,105 @@ compdb(int argc, char *argv[])
...
+			continue;
+
+		for (size_t n = 0; n < e->nout; ++n) {
+			if (e->out[n]->nuse != 0)
+				continue;
+
+			nodes = xreallocarray(nodes, size + 1, sizeof(nodes[0]));
+			nodes[size] = e->out[n];
+			size += 1;
+		}
```
michaelforney: Instead of allocating an array one by one here, then passing the result to `targets_depth`, can't you just replace this for loop with

```c
		targets_depth(e->out, e->nout, depth, 0);
```

?

tool.c
```diff
@@ -221,9 +222,105 @@ compdb(int argc, char *argv[])
...
+		} else {
+			fprintf(stderr, "unknown target tool mode '%s'\n", mode);
+			return 2;
+		}
+	}
+
+	/* Find root nodes (node with 0 use) */
+	size_t size = 0;
+	struct node **nodes = NULL;
+	struct edge *e = NULL;
```
michaelforney: samurai's existing style declares all variables at the start of the function.

michaelforney:
Also, this needs a corresponding update to the manual. If you are not familiar with mdoc(7), see the manual here:
http://mandoc.bsd.lv/man/mdoc.7.html

You can also look at the rest of `samu.1` for examples.

tool.c
Bl4ckb0ne: Done

tool.c
Bl4ckb0ne: Done

tool.c
Bl4ckb0ne: Done

tool.c
Bl4ckb0ne: Done

tool.c
Bl4ckb0ne: I completely missed this but yes, it can be done that way and can avoid reallocating the array. This is done in e8003ad

tool.c
Bl4ckb0ne: This code has been removed in e8003ad.

Bl4ckb0ne:
@michaelforney I fixed the requested changes and wrote some documentation. I'd like to have your feedback on the documentation please.

tool.c
```diff
@@ -221,9 +222,106 @@ compdb(int argc, char *argv[])
...
+{
+	struct edge *e;
+
+	for (e = alledges; e; e = e->allnext) {
+		if (argc == 2) {
+			if (e->nin == 0)
+				continue;
+
+			if (!e->in[0]->gen)
+				printf("%s\n", e->in[0]->path->s);
```
michaelforney: I think this also should loop over all inputs, not just the first one.

tool.c
```diff
@@ -221,9 +222,106 @@ compdb(int argc, char *argv[])
...
+		} else if (argc == 3) {
+			if (e->nout == 0)
+				continue;
```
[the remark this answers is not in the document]
Bl4ckb0ne: Done

tool.c
```diff
@@ -221,9 +222,106 @@ compdb(int argc, char *argv[])
...
+	}
+}
+
+static void
+targets_all(void)
+{
+	struct edge *e = NULL;
+	for (e = alledges; e; e = e->allnext) {
+		if (e->nout == 0)
+			continue;
```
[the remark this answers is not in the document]
Bl4ckb0ne: Done

tool.c
```diff
@@ -221,9 +222,106 @@ compdb(int argc, char *argv[])
...
+
+		for (size_t j = 0; j < indent; ++j)
+			printf("  ");
+
+		printf("%s: %s\n", nodes[i]->path->s, nodes[i]->gen->rule->name);
+
+		if (depth > 1 || depth <= 0)
+			targets_depth(nodes[i]->gen->in, nodes[i]->gen->nin,
+					depth - 1, indent + 1);
```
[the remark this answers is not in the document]
Bl4ckb0ne: Done

tool.c
```diff
@@ -221,9 +222,106 @@ compdb(int argc, char *argv[])
...
+		} else {
+			fprintf(stderr, "unknown target tool mode '%s'\n", mode);
+			return 2;
+		}
+	}
+
+	struct edge *e = NULL;
+	for (e = alledges; e; e = e->allnext) {
+		if (e->nout == 0)
+			continue;
```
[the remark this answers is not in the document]
Bl4ckb0ne: Done

samu.1
```diff
@@ -24,8 +24,13 @@
 .Op Fl C Ar dir
 .Op Fl f Ar buildfile
 .Fl t Cm compdb
-.Op Fl x
```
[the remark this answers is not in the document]
Bl4ckb0ne: Seems like it, I put it back,

tool.c
Bl4ckb0ne: Done

Bl4ckb0ne:
@michaelforney changes are done, I'll let you verify.

tool.c
```diff
@@ -303,11 +297,9 @@ targets(int argc, char *argv[])
...
 	struct edge *e = NULL;
+	size_t n;
 	for (e = alledges; e; e = e->allnext) {
-		if (e->nout == 0)
-			continue;
-
-		for (size_t n = 0; n < e->nout; ++n) {
+		for (n = 0; n < e->nout; ++n) {
 			if (e->out[n]->nuse != 0)
 				continue;
```
michaelforney: This loop didn't seem quite right to me. For every output that is not used by any other action, `targets_depth` is called for the entire list of outputs.

I think this would cause redundancy if there are multiple outputs, and if there were some that were used by other actions, they would get printed anyway.

I pushed a commit which restructures `targets_depth` slightly, moving the loop to surround the recursive call, instead of looping at the toplevel. I believe it should resolve those issues.

tool.c
```diff
@@ -221,9 +222,99 @@ compdb(int argc, char *argv[])
...
+		const char *mode = argv[1];
+		if (strcmp(mode, "rule") == 0) {
+			targets_rule(argc, argv);
+			return 0;
+		} else if (strcmp(mode, "depth") == 0) {
+			if (argc > 1) {
+				char *end;
+				depth = (size_t)strtol(argv[2], &end, 10);
+				if (*end)
+					fprintf(stderr, "depth not valid");
```
michaelforney: This should error out if the depth was invalid.

Also, the format string is missing a trailing newline.

michaelforney:
I pushed a few more commits to your branch. Let me know if they look okay to you.

I want to make a few more changes to the manual, but other than that, I think this is good to go.

Bl4ckb0ne:
Your commits seems fine to me, what changes would you like to make to the manual?

michaelforney:
Pushed now. Thanks again for implementing this!

PR #44, opened 2020-03-04
print status after job like ninja, instead of before

E5ten:
This seems to work correctly but probably has some inadvertent issues I'll need to fix.
EDIT: not noticing any problems, so removing WIP from the title.

### DIFF
--- a/build.c
+++ b/build.c
@@ -238,7 +238,7 @@ jobstart(struct job *j, struct edge *e)
 	extern char **environ;
 	size_t i;
 	struct node *n;
-	struct string *rspfile, *content, *description;
+	struct string *rspfile, *content;
 	int fd[2];
 	posix_spawn_file_actions_t actions;
 	char *argv[] = {"/bin/sh", "-c", NULL, NULL};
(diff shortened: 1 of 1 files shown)


E5ten:
ping?

michaelforney:
Sorry for the delay.

Looking more closely at ninja's code, I think it might actually be more complicated than that:
- If the job uses the console, or the terminal is smart, the status is printed before the job.
- Otherwise, the status is printed after the job, but in the status calculation, `%r` counts the newly finished job as a running job.

I don't quite remember the reason we were looking into this. Something about a missing or extra newline in the output? If that's the case, then perhaps it is not related to this after all (since with a smart terminal, ninja prints the status before the job runs, like samurai).

E5ten:
Can you tell me a range of line numbers and a file from ninja's codebase to look at where this happens? (not sure what to do until I properly get what ninja's doing)

michaelforney:
Sure:

https://github.com/ninja-build/ninja/blob/1f3cc7b7207309c441ce5bf120b57081bbea0c18/src/build.cc#L105-L106
https://github.com/ninja-build/ninja/blob/1f3cc7b7207309c441ce5bf120b57081bbea0c18/src/build.cc#L132-L133
https://github.com/ninja-build/ninja/blob/1f3cc7b7207309c441ce5bf120b57081bbea0c18/src/build.cc#L233-L235

PR #46, opened 2020-03-21
Free old root environment, delete pools, change delevalstr argument type.

angularorbit:
The new `deltree` function header, requires `del` function to have a generic argument type. So I had to change the delevalstr argument type to a `void *`

### DIFF
--- a/env.c
+++ b/env.c
@@ -18,12 +18,22 @@ struct pool consolepool = {.name = "console", .maxjobs = 1};
 static struct treenode *pools;
 
 static void addpool(struct pool *);
+static void delpool(void *);
+static void delrule(void *);
 
 void
 envinit(void)
 {
-	/* TODO: delete old root environment and pools (in case we rebuilt
-	 * build.ninja). for now, we leak memory. */
+	/* free old rootenv and pools in case we rebuilt the manifest
+	 * TODO: some memory is still leaked if subninja is used.
+	 * fixing this would probably require keeping a list of used
+	 * environments and traversing them here. */
+	if (rootenv) {
+		deltree(rootenv->bindings, free);
+		deltree(rootenv->rules, delrule);
+		deltree(pools, delpool);
+		free(rootenv);
+	}
 	rootenv = mkenv(NULL);
 	envaddrule(rootenv, &phonyrule);
 	pools = NULL;
@@ -178,6 +188,16 @@ mkrule(char *name)
 	return r;
 }
 
+static void
+delrule(void *ptr)
+{
+	struct rule *r = ptr;
+	if (r == &phonyrule)
+		return;
+	deltree(r->bindings, delevalstr);
+	free(r);
+}
+
 void
 ruleaddvar(struct rule *r, char *var, struct evalstring *val)
 {
@@ -223,6 +243,15 @@ addpool(struct pool *p)
 {
 	if (treeinsert(&pools, p->name, p))
 		fatal("pool '%s' redefined", p->name);
+}
+
+static void
+delpool(void *ptr)
+{	
+	struct pool *p = ptr;
+	if(p == &consolepool)
+		return;
+	free(p);
 }
 
 struct pool *
--- a/env.c
+++ b/env.c
@@ -29,9 +29,9 @@ envinit(void)
 	 * fixing this would probably require keeping a list of used
 	 * environments and traversing them here. */
 	if (rootenv) {
-		deltree(rootenv->bindings, free);
-		deltree(rootenv->rules, delrule);
-		deltree(pools, delpool);
+		deltree(rootenv->bindings, free, free);
+		deltree(rootenv->rules, NULL, delrule);
+		deltree(pools, NULL, delpool);
 		free(rootenv);
 	}
 	rootenv = mkenv(NULL);
@@ -194,7 +194,8 @@ delrule(void *ptr)
 	struct rule *r = ptr;
 	if (r == &phonyrule)
 		return;
-	deltree(r->bindings, delevalstr);
+	free(r->name);
+	deltree(r->bindings, free, delevalstr);
 	free(r);
 }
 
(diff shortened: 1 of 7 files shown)


env.c
```diff
@@ -19,11 +19,19 @@ static struct treenode *pools;
...
+static void delrule(void *);
+
 void
 envinit(void)
 {
-	/* TODO: delete old root environment and pools (in case we rebuilt
-	 * build.ninja). for now, we leak memory. */
+	/* free old rootenv in case rebuilt the manifest */
+	if (rootenv) {
+		deltree(rootenv->bindings, free);
```
angularorbit: rootenv->bindings values are strings, `string struct` is freed using free.

env.c
```diff
@@ -178,6 +186,19 @@ mkrule(char *name)
 	return r;
 }
 
+static void
+delrule(void *p)
+{
+	struct rule *r = p;
+	
+	if(!r)
+		return;
```
michaelforney: Is it ever the case that we have a NULL rule here? I'd prefer to remove this check if it is unneeded.

env.c
```diff
@@ -19,11 +19,19 @@ static struct treenode *pools;
...
 {
-	/* TODO: delete old root environment and pools (in case we rebuilt
-	 * build.ninja). for now, we leak memory. */
+	/* free old rootenv in case rebuilt the manifest */
+	if (rootenv) {
+		deltree(rootenv->bindings, free);
+		deltree(rootenv->rules, delrule);
+		deltree(pools, delpool);
+		free(rootenv);
+	}
```
michaelforney: I think we still leak memory if `subninja` was used.

Fixing this would probably involve keeping a list of all environments, and traversing the list here. What do you think?

env.h
```diff
@@ -16,21 +16,30 @@ struct pool {
...
 struct rule *mkrule(char *);
+/* add to a rule a variable and its value */
 void ruleaddvar(struct rule *, char *, struct evalstring *);
 
 /* create a new pool with the given name */
 struct pool *mkpool(char *);
 /* lookup a pool by name. fails if it does not exist */
 struct pool *poolget(char *);
 
+/* return an edge's variable evaluated, optionally shell-escaped */
```
michaelforney: These are all fine (thanks!), but could you move these comment changes to a separate commit?

util.c
```diff
@@ -135,8 +135,9 @@ mkstr(size_t n)
 }
 
 void
-delevalstr(struct evalstring *str)
+delevalstr(void *pn)
```
michaelforney: I think we should come up with a consistent  name for this parameter. Perhaps `ptr` or `data`?

env.c
```diff
@@ -19,11 +19,19 @@ static struct treenode *pools;
 
 static void addpool(struct pool *);
```
michaelforney: This blank line can be removed, just group all the forward declarations together.

env.c
angularorbit: I haven't read the whole code base, so I don't for sure. probably unneeded, so I'll remove both this check and the rule-name check, since `mkrule` allocates a name.

util.c
angularorbit: I'll use ptr

env.c
angularorbit: Fair enough.

env.c
angularorbit: I haven't read the whole codebase yet, so I don't know which environments I need to remove. Also no idea what `subninja` is. I'll write a TODO for the extra leaked memory, but keep the code as is.

michaelforney: The root environment that contains definitions in `build.ninja`. If `build.ninja` contains any [`subninja`](https://ninja-build.org/manual.html#ref_scope) lines, new environments will be created with the root environment as a parent, and so on.

Since we only have pointers back to the parent, there is currently no way to traverse the tree of environments. I can fix this in a follow-up commit. A TODO is fine for now.

env.c
michaelforney: The rule name check should not be needed either.

I believe the same should be true for pools as well.

env.h
angularorbit: Ok, done!

angularorbit:
How's it looking?

env.c
angularorbit: just edited the pools. was experimenting with git rebase, and it merged the amended commit smoothly.

angularorbit: Also after I had split the comment into 2 and pushed the first, I had lost my comments, but was able to find them with `git reflog`. That's next level!

michaelforney: Thanks. Yeah, `git rebase` and `git reflog` are very handy tools, indeed.

michaelforney:
Looking great, thanks!

I was doing some testing with valgrind, and found a couple of issues:
- The `phonyrule` and `consolepool` should not be freed, since they are globals. I think we could just have `delrule` and `delpool` only free when it is not one of those.
- Since the rule and pool names are also used as the tree key, we end up freeing those twice. I think we can solve this by skipping the name free in `delrule` and `delpool`. If we did it the other way and remove the key free, then environment and rule variable names wouldn't get freed.

angularorbit:
Rebased again. We need a taste suite. Also I need to start using a debugger.

michaelforney:
Looks like there is still an issue, since string literals for "console" and "phony" are freed. Best thing I can think of is to pass separate functions for freeing the key and the value, so I pushed a commit to your branch that does that.

Does this seem okay to you?

michaelforney:
Another possibility is to not actually insert the `phonyrule` and `consolepool` into the tree, just have a special case for them in `envrule` and `poolget`.

angularorbit:
I think passing a tree key-free function makes sense, because `tree` is not responsible for allocating the keys, and shouldn't be responsible for freeing them.

Whether to remove `phonyrule` and `consolepool` from the tree, and special cases, I don't know. I'll leave the decision to you.

michaelforney:
Whoops, I pushed the change before I read your edit. I don't understand what you mean about which parts of the tree need to be freed and which don't. It just depends on whether the pool/rule was a special static one.

Feel free to send another PR if you think we should use the other approach.

Issue #31, opened 2020-01-10
Samurai fails to build https://github.com/bvschaik/julius


ibara:
Hello --

On an OpenBSD/amd64 machine, samurai fails to build the repository at https://github.com/bvschaik/julius

Samurai prints the line
`[1/2] /usr/local/bin/cmake -P /home/brian/julius/build/CMakeFiles/VerifyGlobs.cmake`
100 times, then prints
`samu: manifest 'build.ninja' dirty after 100 tries`

In contrast, Ninja prints
`[0/2] /usr/local/bin/cmake -P /home/brian/julius/build/CMakeFiles/VerifyGlobs.cmake`
then proceeds to build the repository.

Let me know if you need anything else.

michaelforney:
Thanks for reporting this issue. The best way to debug this sort of problem is to use `-d explain`.

```
samu: explain /tmp/julius/build/CMakeFiles/VerifyGlobs.cmake_force: phony and no inputs
samu: explain /tmp/julius/build/CMakeFiles/cmake.verify_globs: input is dirty
samu: explain build.ninja: input is dirty
[1/2] /usr/bin/cmake -P /tmp/julius/build/CMakeFiles/VerifyGlobs.cmake
samu: explain /tmp/julius/build/CMakeFiles/VerifyGlobs.cmake_force: phony and no inputs
samu: explain /tmp/julius/build/CMakeFiles/cmake.verify_globs: input is dirty
samu: explain build.ninja: input is dirty
[1/2] /usr/bin/cmake -P /tmp/julius/build/CMakeFiles/VerifyGlobs.cmake
...
```

```
ninja explain: output /tmp/julius/build/CMakeFiles/VerifyGlobs.cmake_force of phony edge with no inputs doesn't exist
ninja explain: /tmp/julius/build/CMakeFiles/VerifyGlobs.cmake_force is dirty
ninja explain: /tmp/julius/build/CMakeFiles/cmake.verify_globs is dirty
[0/2] /usr/bin/cmake -P /tmp/julius/build/CMakeFiles/VerifyGlobs.cmake
ninja explain: deps for 'CMakeFiles/julius.dir/src/platform/arguments.c.o' are missing
ninja explain: CMakeFiles/julius.dir/src/platform/arguments.c.o is dirty
...
[...]

PR #48, opened 2020-04-05
dyndep support

### LINKED ISSUE #36: Implement dyndeps
This is the main new feature in ninja 1.10.

Documentation is here:
https://github.com/ninja-build/ninja/blob/master/doc/manual.asciidoc#ref_dyndep

Duncaen:
```ninja
rule untar
  command = tar xf $in && touch $out
rule scantar
  command = printf "ninja_dyndep_version = 1\nbuild %s | %s: dyndep\n  restat = true\n" "$stamp" $
  "$$(tar tf $in | tr "\n" " ")" >$out.tmp && mv $out.tmp $out
rule cat
  command = cat $in >$out.tmp && mv $out.tmp $out
build foo.tar.dd: scantar foo.tar
  stamp = foo.tar.stamp
build foo.tar.stamp: untar foo.tar || foo.tar.dd
  dyndep = foo.tar.dd
```

```
samurai$ tar tf test/foo.tar
foo
bar
```
```
samurai$ ./samu -C test/ -d explain
samu: explain foo.tar.dd: missing
samu: explain foo.tar.stamp: missing
[1/2] printf "ninja_dyndep_version = 1\nbuild %s | %s: dyndep\n  restat = true\n" "foo.tar.stamp" "$(tar tf foo.tar | tr "\n" " ")" >foo.tar.dd.tmp && mv foo.tar.dd.tmp foo.tar.dd
samu: loading dyndep file: 'foo.tar.dd'
[2/2] tar xf foo.tar && touch foo.tar.stamp
```
```
samurai$ ./samu -C test/ -d explain
samu: loading dyndep file: 'foo.tar.dd'
samu: nothing to do
```
```
samurai$ rm test/foo
samurai$ ./samu -C test/ -d explain
samu: loading dyndep file: 'foo.tar.dd'
samu: explain foo: missing
samu: explain foo.tar.stamp: output of generating action is dirty
samu: explain bar: output of generating action is dirty
[1/1] tar xf foo.tar && touch foo.tar.stamp
```
```
samurai$ ./samu -C test/ -d explain -t clean
samu: loading dyndep file: 'foo.tar.dd'
remove foo.tar.stamp
remove foo
remove bar
remove foo.tar.dd
```

### DIFF
--- a/build.c
+++ b/build.c
@@ -106,6 +106,10 @@ queue(struct edge *e)
 {
 	struct edge **front = &work;
 
+	if (e->flags & FLAG_QUEUED)
+		return;
+	e->flags |= FLAG_QUEUED;
+
 	if (e->pool && e->rule != &phonyrule) {
 		if (e->pool->numjobs == e->pool->maxjobs)
 			front = &e->pool->work;
--- a/build.c
+++ b/build.c
@@ -115,13 +115,43 @@ queue(struct edge *e)
 	*front = e;
 }
 
+static void
+computedirty(struct edge *e, struct node *newest)
+{
+	struct node *n;
+	size_t i;
+	bool generator, restat;
+
+	/* all outputs are dirty if any are older than the newest input */
+	generator = edgevarbool(e, "generator");
+	restat = edgevarbool(e, "restat");
+	for (i = 0; i < e->nout && !(e->flags & FLAG_DIRTY_OUT); ++i) {
+		n = e->out[i];
+		if (isdirty(n, newest, generator, restat)) {
+			n->dirty = true;
+			e->flags |= FLAG_DIRTY_OUT;
+		}
+	}
+	if (e->flags & FLAG_DIRTY) {
+		for (i = 0; i < e->nout; ++i) {
+			n = e->out[i];
+			if (buildopts.explain && !n->dirty) {
+				if (e->flags & FLAG_DIRTY_IN)
+					warn("explain %s: input is dirty", n->path->s);
+				else if (e->flags & FLAG_DIRTY_OUT)
+					warn("explain %s: output of generating action is dirty", n->path->s);
+			}
+			n->dirty = true;
+		}
+	}
+}
+
 void
 buildadd(struct node *n)
 {
 	struct edge *e;
 	struct node *newest;
 	size_t i;
-	bool generator, restat;
 
 	e = n->gen;
 	if (!e) {
@@ -158,28 +188,7 @@ buildadd(struct node *n)
 		if (n->dirty || (n->gen && n->gen->nblock > 0))
 			++e->nblock;
 	}
-	/* all outputs are dirty if any are older than the newest input */
-	generator = edgevar(e, "generator", true);
-	restat = edgevar(e, "restat", true);
-	for (i = 0; i < e->nout && !(e->flags & FLAG_DIRTY_OUT); ++i) {
-		n = e->out[i];
-		if (isdirty(n, newest, generator, restat)) {
-			n->dirty = true;
-			e->flags |= FLAG_DIRTY_OUT;
-		}
-	}
-	if (e->flags & FLAG_DIRTY) {
-		for (i = 0; i < e->nout; ++i) {
-			n = e->out[i];
-			if (buildopts.explain && !n->dirty) {
-				if (e->flags & FLAG_DIRTY_IN)
-					warn("explain %s: input is dirty", n->path->s);
-				else if (e->flags & FLAG_DIRTY_OUT)
-					warn("explain %s: output of generating action is dirty", n->path->s);
-			}
-			n->dirty = true;
-		}
-	}
+	computedirty(e, newest);
 	if (!(e->flags & FLAG_DIRTY_OUT))
 		e->nprune = e->nblock;
 	if (e->flags & FLAG_DIRTY) {
--- a/build.c
+++ b/build.c
@@ -443,6 +443,8 @@ edgedone(struct edge *e)
 	bool restat, prune, pruned;
 	int64_t old;
 
+	/* mark edge clean for dyndepload */
+	e->flags &= ~FLAG_DIRTY;
 
 	newest = NULL;
 	prune = pruned = false;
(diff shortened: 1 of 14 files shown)


Duncaen:
One big open issue is that it currently does not recompute the newly added dependencies when they are loaded after the dyndep node is done.

dyndep.c
```diff
@@ -0,0 +1,161 @@
...
+		name = scanname(s);
+		if (strcmp(name, "restat") != 0)
+			fatal("unexpected variable '%s' in dyndep", name);
+		parselet(s, &str);
+		envaddvar(e->env, "restat", enveval(env, str));
+	}
+}
+
+static void
+dyndepparse(struct node *n)
```
michaelforney, line 107: I think this should just take the file path directly.

graph.h
```diff
@@ -17,8 +17,7 @@ struct node {
 
 	/* generating edge and dependent edges.
 	 *
-	 * only gen and nuse are set in parse.c:parseedge; use is allocated and
-	 * populated in build.c:computedirty. */
+	 * gen and use are set in parse.c:parseedge */
```
michaelforney, line 20: I think you can just remove this part of the comment completely. Now that `gen` and `use` are set during edge construction with everything else, there is no need to clarify here.

deps.c
```diff
@@ -300,7 +300,7 @@ depsparse(const char *name, bool allowmissing)
 {
 	static struct buffer buf;
 	static struct nodearray deps;
-	static size_t depscap;
+	static size_t depscap = 0;
```
michaelforney, line 303: It's not necessary to initialize static variables to 0.

michaelforney (commented):
Thank you so much for working on this! This is an important feature for samurai to keep up with current ninja, and I'm not sure when I would've gotten around to implementing it.

My first impression after skimming these changes is that this looks great. I really appreciate you taking the care to match the existing style and design of samurai. I haven't read up on the details of dyndeps yet, so I'll do that and get back to you with a more thorough review.

Duncaen:
Ok nice, I opened the PR early to get some initial feedback and to avoid duplicating work.

I'm going to fix some outstanding issues:

* recompute `nblock`/`nprune` after loading the dyndep.
* some cases like `-t clean` should be able to delete broken dyndep files and not call `fatal` if the file can't be parsed.
* write down some details not clear form the manual.
* come up with more extensive tests and maybe use some [ninja files on github](https://github.com/search?l=ninja&q=dyndep&type=Code) for testing.
* probably a few more after more testing

Duncaen:
> * some cases like `-t clean` should be able to delete broken dyndep files and not call `fatal` if the file can't be parsed.

This would either require to add error return values and a lot of ifs to the current scanner which makes everything more complicated.
Would using `setjmp` be acceptable here so that instead of calling fatal directly, it could be handled whether it should be fatal or not by setting a jump buffer. 

Current issues:

* `ntotal` is not updated correctly.
* need to rewrite `checkversion` to compare the numbers: comparing `1.5.0` to `1.10.0` fails.
* cmake tests for `cleandead`, `recompact` and `restat` tools when building fortran modules with dyndep support; while testing I've just added stubs that are always successful.

michaelforney:
> > * some cases like `-t clean` should be able to delete broken dyndep files and not call `fatal` if the file can't be parsed.
> 
> This would either require to add error return values and a lot of ifs to the current scanner which makes everything more complicated.
> Would using `setjmp` be acceptable here so that instead of calling fatal directly, it could be handled whether it should be fatal or not by setting a jump buffer.

Hmm... yes, I see the problem.

I think I'd like to make the scanner be able to report errors instead of using `setjmp`. Yes, it would add a bit of complexity to the parser, but I don't think it would be too bad.

> Current issues:
> 
>     * `ntotal` is not updated correctly.
> 
>     * need to rewrite `checkversion` to compare the numbers: comparing `1.5.0` to `1.10.0` fails.

Looks like these two are resolved now, is that right?

>     * cmake tests for `cleandead`, `recompact` and `restat` tools when building fortran modules with dyndep support; while testing I've just added stubs that are always successful.

Do you know why it needs those subtools?

build.c
```diff
@@ -146,6 +146,49 @@ computedirty(struct edge *e, struct node *newest)
 	}
 }
 
+void
+buildupdate(struct node *n)
```
michaelforney, line 155: I wonder if we could reuse even more code here by passing input and output nodes as parameters to a helper function

build.c
```diff
@@ -106,6 +106,10 @@ queue(struct edge *e)
 {
 	struct edge **front = &work;
 
+	if (e->flags & FLAG_QUEUED)
+		return;
+	e->flags |= FLAG_QUEUED;
```
michaelforney, line 111: Could you explain why we would otherwise queue edges multiple times with dyndeps?

build.c
```diff
@@ -443,6 +443,8 @@ edgedone(struct edge *e)
 	bool restat, prune, pruned;
 	int64_t old;
 
+	/* mark edge clean for dyndepload */
+	e->flags &= ~FLAG_DIRTY;
```
michaelforney, line 452: I think this might break manifest rebuilds. We currently use the dirty flag to check if we actually needed to rebuild it (see #31).

build.c
```diff
@@ -115,13 +115,43 @@ queue(struct edge *e)
...
+static void
+computedirty(struct edge *e, struct node *newest)
+{
+	struct node *n;
+	size_t i;
+	bool generator, restat;
+
+	/* all outputs are dirty if any are older than the newest input */
+	generator = edgevarbool(e, "generator");
+	restat = edgevarbool(e, "restat");
```
michaelforney, line 132: Just checking, does ninja treat set-but-empty `generator` and `restat` as false?

michaelforney (commented):
Thanks for factoring this into smaller changes. It makes it a lot easier to review.

But, could you expand on the commit descriptions a bit? Explaining the problems it solves, or how it will be used in future commit, etc.

In particular, I don't quite understand "build: edgedone: prune all output nodes if one is being pruned" and "log: create nodes if they don't exist so dyndeps can use logmtime".

The pruning stuff is probably the trickiest part of samurai, so I want to make sure we don't accidentally introduce any regressions.

orbea:
@Duncaen & @michaelforney Any news on the status of this feature?

ismaell:
Was this superseded by some other PR?

PR #50, opened 2020-05-03
[WIP] Load average

Salamandar:
That's a complete WIP. I'll have parallel PRs too for the first unrelated commits :)

I found out that PR will be a bit more difficult than expected :
* If I can't start any jobs at first, Samu quits (Ninja starts max_jobs without checking the load…)
* I'll have to add an item in `fds` to notify `poll` if load gets under an acceptable value (Ninja waits for a job to stop).

Feel free to give me any suggestions.


michaelforney:
Thanks for getting started with this.

Sorry, I don't have the time right now to dig in to how ninja deals with these problems, so I don't have any real actionable suggestions.

Regarding `poll`, if ninja just waits for a job to stop (I'm assuming you mean `wait` or `waitpid` here), how does it deal with the possibility of the pipes getting full? Wouldn't that cause a potential deadlock where the jobs can't finish because they are blocked on writing to the pipe?

Salamandar:
OK, I finally pulled it off.
I'm using a secondary thread that writes in a file descriptor, to be handled by the `poll` that waits for jobs to finish.
The thread also polls on the writing side of the fd to listen to `close`, with a timeout of 1s.

> Regarding poll, if ninja just waits for a job to stop (I'm assuming you mean wait or waitpid here), how does it deal with the possibility of the pipes getting full? Wouldn't that cause a potential deadlock where the jobs can't finish because they are blocked on writing to the pipe?

Uhm, I honestly don't know.

Salamandar:
@michaelforney What do you think about this PR ?

michaelforney:
I don't want to use threads to solve this problem. There must be another approach.

As far as I can tell, ninja also polls indefinitely when waiting for a job to finish. I agree that it would be nice to start a new job when the load average drops under the limit, but this does not appear to be what ninja does.

If we do want to differ from ninja in this respect, isn't a timeout to the existing `poll()` sufficient?

michaelforney:
Can you rebase to remove the unrelated commits from this PR?

Also, could you try adding a timeout to the existing `poll()` instead of a separate load average thread?

michaelforney:
Closing in favor of #82.

PR #51, opened 2020-05-03
Add pragma once to every header (allows unity build)

Salamandar:
That's pretty straightforward : Adding `#pragma once` to every header allows unity build (compile all files together to allow better optimization).

### DIFF
--- a/arg.h
+++ b/arg.h
@@ -1,3 +1,5 @@
+#pragma once
+
 extern const char *argv0;
 
 #define ARGBEGIN \
--- a/build.h
+++ b/build.h
@@ -1,3 +1,5 @@
+#pragma once
+
 struct node;
 
 struct buildoptions {
--- a/deps.h
+++ b/deps.h
@@ -1,3 +1,5 @@
+#pragma once
+
 struct edge;
 
 void depsinit(const char *);
--- a/env.h
+++ b/env.h
@@ -1,3 +1,5 @@
+#pragma once
+
 struct evalstring;
 struct string;
 
--- a/graph.h
+++ b/graph.h
@@ -1,3 +1,5 @@
+#pragma once
+
 #include <stdint.h>  /* for uint64_t */
 
 /* set in the tv_nsec field of a node's mtime */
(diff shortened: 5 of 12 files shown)


E5ten:
Since `#pragma once` is an extension, it would probably be better to use `#ifndef` guards.

michaelforney:
Yes, samurai is written in ISO C99, so we can't use `#pragma once`. But, I don't think I want to add `#ifndef` guards either.

Are you actually able to detect a difference in performance by compiling all the files together? If so, could you attach the results of your tests?

Salamandar:
> I don't think I want to add #ifndef guards either.
Why ?

> Are you actually able to detect a difference in performance by compiling all the files together? If so, could you attach the results of your tests?

Actually, yesterday I didn't have a big enough bench set, and I was wondering if you had one.

michaelforney:
> > I don't think I want to add #ifndef guards either.
> 
> Why ?

A simple way to prevent the need for include guards is to not include headers in other headers. See the "Include files" section of http://doc.cat-v.org/bell_labs/pikestyle.

> > Are you actually able to detect a difference in performance by compiling all the files together? If so, could you attach the results of your tests?
> 
> Actually, yesterday I didn't have a big enough bench set, and I was wondering if you had one.

I usually just do no-op builds in large projects like chromium or llvm. In any build that runs any commands, samurai's contribution to the build time is lost in the noise.

Salamandar:
Well I just did a benchmark on Gimp, and the difference is small but visible : 
```
no unity: 0.12967142857142855
unity:    0.12873913043478247
```

michaelforney:
What does "count" mean here? Is that the number of samples you ran?

Note that you can build with `LDFLAGS=-flto` to enable LTO, you don't need meson.

Salamandar:
Yes, `count` is the number of samples. I did a much longer run (8000 samples) with the same result.

michaelforney:
Sorry, I'm not interested in supporting a unity build. If you really want to do this, you can just cat the files after removing their `#include "..."` lines, and prepending some initial includes:

```
{ printf '#include <%s>\n' stdio.h; printf '#include "%s"\n' *.h; grep -hv '^#include "' *.c; } | gcc -x c - -o samu
```

PR #52, opened 2020-05-03
Meson

Salamandar:
I added Meson build system : 
* correct header dependencies
* No Make/GCC/Linux tree dependency (road to Windows compat)
* Add shell completion definitions from Ninja (License kept and mention added)

I could rewrite the ninja.build or Makefile to transparently use Meson (or not touch it to keep Meson optional).


michaelforney:
Thanks, but as I explained in the previous PR, I'm not interested in adding the meson build system to samurai. Could you explain what problems you are having with the Makefile, and how they are solved with meson?

> * correct header dependencies

The Makefile declares dependencies on all headers for every object. A complete build takes a fraction of a second, so even when a header changes, there is not much time wasted.

> * No Make/GCC/Linux tree dependency (road to Windows compat)

The Makefile only uses POSIX features. It does not depend on Linux or gcc, and should work on any POSIX system. I believe it should even work with nmake on Windows (though I have not tried it, since we still use a few POSIX-specific APIs in the source).

> * Add shell completion definitions from Ninja (License kept and mention added)

This seems unrelated, though I might be open to adding them in a `contrib` directory. But I think we should write new ones for samurai, not copy from ninja. For example, samurai does not support `-d list` or `-t list`, so I'm not sure it would even work.

Salamandar:
Crap, I was almost sure I already did a PR but couldn't find it yesterday :(

Although, I have a bit more arguments now.

> The Makefile only uses POSIX features. It does not depend on Linux or gcc, and should work on any POSIX system. I believe it should even work with nmake on Windows (though I have not tried it, since we still use a few POSIX-specific APIs in the source).

True, but you are defining install directories as `prefix/bin`, `prefix/man1` and that may not be correct on windows.
Also, porting Samurai on Windows may need some other checks ("what OS are we on", "is this function/header available") and that may be difficult (and verbose) to add them to the Makefile.
For example, my "load monitoring" PR uses pthread, that is not available on Windows (well the API is but no additional link is needed, IIRC).

Meson also allows to build debug/release/LTO/… very easily.

Salamandar:
> This seems unrelated, though I might be open to adding them in a contrib directory. But I think we should write new ones for samurai, not copy from ninja. For example, samurai does not support -d list or -t list, so I'm not sure it would even work.

Yes, you're right, I shouldn't have done a single PR for both.
For "user-friendliness", maybe samurai should support `-d list` and `-t list` ?

michaelforney:
> True, but you are defining install directories as `prefix/bin`, `prefix/man1` and that may not be correct on windows.
> Also, porting Samurai on Windows may need some other checks ("what OS are we on", "is this function/header available") and that may be difficult (and verbose) to add them to the Makefile.
> For example, my "load monitoring" PR uses pthread, that is not available on Windows (well the API is but no additional link is needed, IIRC).

Possibly, but it should also be pretty easy to copy the executable to wherever it needs to be. You don't need to build the install target.

Anyway, it's not worth worrying about this now. We can revisit this if someday samurai works on Windows and the Makefile proves to be insufficient.

> Meson also allows to build debug/release/LTO/… very easily.

As does the Makefile, just set CFLAGS/LDFLAGS to whatever you want in your environment or on the command-line.

michaelforney:
> Yes, you're right, I shouldn't have done a single PR for both.
> For "user-friendliness", maybe samurai should support `-d list` and `-t list` ?

Sure, I'm fine with adding this.

PR #57, opened 2020-10-23
Implement more tools as well as -n

bonzini:
This pull request brings a few more ninja features to samurai, namely:
* `-t list`
* `-t query` (QEMU uses it to bridge Make with ninja)
* dry-run (`-n`)
* `-t commands`, which is effectively a variation on dry-run

### DIFF
--- a/tool.c
+++ b/tool.c
@@ -7,6 +7,7 @@
 #include <string.h>
 #include <unistd.h>
 #include "arg.h"
+#include "build.h"
 #include "env.h"
 #include "graph.h"
 #include "tool.h"
@@ -123,6 +124,45 @@ clean(int argc, char *argv[])
 		}
 	}
 
+	return ret;
+}
+
+/* Depth-first search.  */
+static void
+targetcommands(struct node *n)
+{
+	struct edge *e = n->gen;
+	struct string *command;
+	size_t i;
+	if (!e || (e->flags & FLAG_WORK))
+		return;
+	e->flags |= FLAG_WORK;
+	for (i = 0; i < e->nin; ++i)
+		targetcommands(e->in[i]);
+	command = edgevar(e, "command", true);
+	if (command && command->n)
+		puts(command->s);
+}
+
+static int
+commands(int argc, char *argv[])
+{
+	int ret = 0;
+	struct node *n;
+
+	if (argc > 1) {
+		while (*++argv) {
+			n = nodeget(*argv, 0);
+			if (!n) {
+				warn("unknown target '%s'", *argv);
+				ret = 1;
+				continue;
+			}
+			targetcommands(n);
+		}
+	} else {
+		dodefault(targetcommands);
+	}
 	return ret;
 }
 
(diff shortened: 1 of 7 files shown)


michaelforney:
Thanks for the PR! I applied the first two with some minor stylistic tweaks.

The dry-run mode also looks good, but I think the `commands` tool needs some more work. Based on my experiments with ninja, it should not be skipping commands for objects that are already up to date, so maybe reusing the build.c code will not work. Instead, maybe it could just walk through the graph, printing the command of each edge (setting an the appropriate edge flag to avoid duplication)

bonzini:
In theory that was the purpose of the

```
-		n->dirty = false;
+		n->dirty = (buildopts.mode == COMMANDS);
```

change but yes it only works on my silly testcase. I was hoping to skip the duplication of the graph walk but it seems simpler indeed to do that.

bonzini:
I updated the dry-run patch to use a bool instead of an enum, since the mode was mostly useful for the `commands` tool.

tool.c
```diff
@@ -126,6 +127,48 @@ clean(int argc, char *argv[])
...
+/* Depth-first search.  */
+static void
+targetcommands(struct node *n)
+{
+	struct edge *e = n->gen;
+	struct string *command;
+	size_t i;
+	if (!e || (e->flags & FLAG_DIRTY_OUT))
+		return;
+	e->flags |= FLAG_DIRTY_OUT;
```
michaelforney, line 139: Not that it matters, but I'd use `FLAG_WORK` here, since it seems to be closest to what we want here.

michaelforney (commented):
Thanks, this looks good. Just a couple of minor comments.

bonzini:
QEMU 5.2 is going to be out in a month or so and it currently doesn't build on Alpine, so it would b nice to have a new release with at least the "query" tool.

jvinolas:
I tried to compile QEMU 5.2 with samurai-1.1-r1 that has query tool and the build goes beyond, but fails at some point with 'ninja: subcommand not found'.

bonzini:
Can you be more specific? What comes before the "subcommand not found" error?

michaelforney:
Merged now, sorry for the delay. I'll try to get a release out in the next few days (and will be sure to test with QEMU 5.2).

eli-schwartz:
Random comment now I tried this:

```
$ samu -t list
samu: open build.ninja: No such file or directory
```

Works fine if I first `cd` to a directory with a build.ninja that can be built, but this tool is in the odd situation of not needing to care about that (and ninja doesn't error out in the same situation).

michaelforney:
> Works fine if I first cd to a directory with a build.ninja that can be built, but this tool is in the odd situation of not needing to care about that (and ninja doesn't error out in the same situation).

This is now fixed in 3629b72d.

michaelforney:
@bonzini I'm thinking about `-t list` a bit more, and am realizing that it is a bit redundant since the subtools are already documented in the manual. It is a similar case to samurai only having a usage message instead of a full help text with `-h` like ninja does (and no manual). Can you comment on whether `-t list` is used by the qemu build system itself, or is it just purely for convenience?

PR #59, opened 2020-11-12
Fix mtime calculation on SunOS

dhgutteridge:


dhgutteridge:
Because you're asking for POSIX standards, modern Solaris derivates
apply underscores to some time-related structures and members as
"standards namespace safe versions of these definitions".

michaelforney:
Thanks for the PR.

> Because you're asking for POSIX standards, modern Solaris derivates
apply underscores to some time-related structures and members as
"standards namespace safe versions of these definitions".

That logic seems backwards to me. Why does it hide the POSIX members when we ask for POSIX standards?

If you add `-D __EXTENSIONS__` to your CFLAGS, does that work around the issue?

dhgutteridge:
> Thanks for the PR.
> 
> > Because you're asking for POSIX standards, modern Solaris derivates
> > apply underscores to some time-related structures and members as
> > "standards namespace safe versions of these definitions".
> 
> That logic seems backwards to me. Why does it hide the POSIX members when we ask for POSIX standards?

The specific conditions (and comments) are:
[https://github.com/illumos/illumos-gate/blob/4e0c5eff9af325c80994e9527b7cb8b3a1ffd1d4/usr/src/uts/common/sys/stat.h#L42](https://github.com/illumos/illumos-gate/blob/4e0c5eff9af325c80994e9527b7cb8b3a1ffd1d4/usr/src/uts/common/sys/stat.h#L42)
and
[https://github.com/illumos/illumos-gate/blob/4e0c5eff9af325c80994e9527b7cb8b3a1ffd1d4/usr/src/uts/common/sys/time_std_impl.h#L28](https://github.com/illumos/illumos-gate/blob/4e0c5eff9af325c80994e9527b7cb8b3a1ffd1d4/usr/src/uts/common/sys/time_std_impl.h#L28)

I agree, I find a couple of things there counter-intuitive.

> If you add `-D __EXTENSIONS__` to your CFLAGS, does that work around the issue?

Applying __EXTENSIONS__ across the board fixes compilation of graph.c but breaks util.c because reallocarray() gets redefined. I'd suggested what I did in part just because you already have a special case for macOS, so concentrating two special cases expressed in the same place and manner seemed cleaner to me than potentially two different means of handling related special cases, should this get complicated. (Assuming there aren't broader SunOS issues; I haven't noticed any.)

michaelforney:
> I'd suggested what I did in part just because you already have a special case for macOS.

The ifdef for macOS was added with a lot of hesitation; I'm trying really hard to avoid the broken mess that was https://github.com/ninja-build/ninja/blob/a65f0d42dc8820cc4d80561ddcd4b693c0adc892/src/disk_interface.cc#L205-L221. There is no combination of preprocessor defines to get macoOS to define the st_mtim member, and it is not clear where to file a bug about it, or if it would ever get fixed.

I'm very reluctant to add platform-specific workarounds for platforms that don't support POSIX-2008 properly. Perhaps those conditions were added before `st_mtim` was standardized in POSIX-2008. At this point I think they should be removed. Can you send a patch (or at least file a bug) to illumos to remove the conditions that hide st_mtim with _POSIX_C_SOURCE, and then include a reference here? That way we at least have a path toward a proper solution.

dhgutteridge:
I've opened a bug report with illumos here:
[https://www.illumos.org/issues/13327](https://www.illumos.org/issues/13327)

(Looking through their bug tracker, I see there are other POSIX 2008 compliance bugs filed, e.g.,
[https://www.illumos.org/issues/1609](https://www.illumos.org/issues/1609)
and
[https://www.illumos.org/issues/5429](https://www.illumos.org/issues/5429)
which have been open for years. The latter example issue has a slight relationship to ours.)

michaelforney:
Thanks for filing the issue! I merged your commit with a comment linking back to the bug.

PR #60, opened 2020-12-18
Makefile: support VPATH builds by properly using '$<'.

markus-oberhumer:


michaelforney:
The meaning of `$<` is unspecified in target rules. I'd like to stick with standard POSIX make features, if possible.

Which make implementations have you tested this with?

markus-oberhumer:
Indeed - after looking at https://pubs.opengroup.org/onlinepubs/009695399/utilities/make.html it seems you are right, so I'm closing this issue.

(But then, POSIX make does not support VPATH anyway - in fact it looks extremely limited.)

PR #64, opened 2021-03-22
Add warning comment in build.ninja for packagers

### LINKED ISSUE #63: Not the same binary using Makefile or build.ninja in some systems
Hi!                                                                              
                                                                                 
Trying to building samurai using `Makefile` and `build.ninja` in my system result in a different binary in each case.                                              
                                                                                 
After inspect the files, I can confirm that is due to the usage of `$(CC)` and       
`$(CFLAGS)` in the `Makefile`, that depends of the current environment.  These   
variables are *hardcoded* in `build.ninja` to `cc` and ``.                       
                                                                                 
Reading about how to use the environment variables `$CC` and `$CFLAGS` in the   
`build.ninja` file, I found from <https://lwn.net/Articles/706404/>:             
                                                                                 
> If you want to support environment variables like $CFLAGS, it is best practice 
> to read these variables in the configure script, and bake the values into the  
> Ninja file.                                                                    
                                                                                 
It's worth it fix this?                                                          
                                                                                 
If yes, i tried two ideas that works:                                              
                                                                                 
1. Hardcode the same values for `$CC` and `$CFLAGS` in `Makefile` and            
    `build.ninja`:
    ```diff                                                                         
    --- Makefile.orig   2021-03-08 16:40:24.565086354 -0300                      
    +++ Makefile        2021-03-08 16:40:39.953077035 -0300                      
    @@ -1,6 +1,8 @@                                                              
     .POSIX:                                                                     
     .PHONY: all install clean                                                   
                                                                                 
    +CC=cc                                                                       
    +CFLAGS=                                                                     
     PREFIX=/usr/local                                                           
     BINDIR=$(PREFIX)/bin                                                        
     MANDIR=$(PREFIX)/share/man                                                  
    ```                         
    This always works, but don't allow to change the compiler using environment  
    variables.  Sure this is a problem with some distributions.   
2. I removed the file `build.ninja` and create it using the script `configure.sh`:
    ```shell                                                                        
    #!/bin/sh                                                                    
    set -e                                                                       
                                                                                 
    cat > build.ninja <<END                                                      
    cc = ${CC:-cc}                                                               
    cflags = ${CFLAGS} -std=c99 -Wall -Wextra -Wpedantic -Wno-unused-parameter   
    END                                                                          
                                                                                 
    cat >> build.ninja <<'END'                                                   
    (rest of the content of the file build.ninja in the repository).             
    ```   
    This option only works if the `CC` and `CFLAGS` environment variables are    
    defined and adds a step in the build.                                            
    *Note*: In my system (debian bullseye) if `$CFLAGS` is not defined as an     
    environment variable `Makefile` uses the option `-O1`.                       
                                                                                 
Maybe another approach can be create an script (`bootstrap.sh` or                
`configure.sh --bootstrap` a la ninja) that builds `samu` binary without
needing of `make` or `samu`?.  Obviously, the current `Makefile` has instructions   
to install `samu` and the related files in the system.

Regards and thanks for your work!

P.D.: [Build script used to check the generated binaries](https://paste.sr.ht/~jumapico/c2bd56948c0771b715d2c08e94187cf2477b3217)

jumapico:
Closes #63.

### DIFF
--- a/build.ninja
+++ b/build.ninja
@@ -1,3 +1,5 @@
+# Demo file that's build samurai
+# -- Distributions packaging samurai are expected to use the Makefile --
 cc = cc
 cflags = -std=c99 -Wall -Wextra -Wpedantic -Wno-unused-parameter
 


michaelforney:
Thanks for the pull-request, but I just ended up deleting build.ninja instead.

PR #69, opened 2021-04-04
Fix floating point exception

t6:
Samu dies with SIGFPE when NINJA_STATUS="%p [%s/%t]" with a basically
empty build file:

	build all: phony

This happens for example during the gnome-backgrounds build.

Found during a FreeBSD Ports exp-run that uses samurai instead of ninja:

http://package18.nyi.freebsd.org/data/122amd64-default-foo/2021-04-04_07h37m38s/logs/errors/gnome-backgrounds-3.38.0.log

### DIFF
--- a/build.c
+++ b/build.c
@@ -226,7 +226,11 @@ formatstatus(char *buf, size_t len)
 			n = snprintf(buf, len, "%zu", ntotal - nstarted);
 			break;
 		case 'p':
-			n = snprintf(buf, len, "%3zu%%", 100 * nfinished / ntotal);
+			if (ntotal > 0) {
+				n = snprintf(buf, len, "%3zu%%", 100 * nfinished / ntotal);
+			} else {
+				n = snprintf(buf, len, "  0%%");
+			}
 			break;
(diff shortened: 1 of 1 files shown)


michaelforney:
Thanks for the PR. I don't think we even end up printing a status if `ntotal` is 0. I think the problem is that the initial `formatstatus(NULL, 0)` is done to ensure that the status format string is valid, even if we never end up using it.

I pushed a commit to return from `build()` early in that case, which should solve the problem. Can you verify that gnome-backgrounds builds successfully with latest master?

> Found during a FreeBSD Ports exp-run that uses samurai instead of ninja

I'm curious about the other results from this run. I wouldn't be surprised if there are a handful of argument-order bugs (i.e. `samu foo -v`) and maybe some underspecified dependencies, but I'm interested to know if there are any other issues you've hit.

t6:
Thanks for fixing this so quickly. Yes, gnome-background builds fine with the latest master. I'll let you know if we run into more problems or how the exp-run went overall.

PR #74, opened 2021-04-28
build: Add DOCDIR.

orbea:
This adds `DOCDIR` and `DATADIR` which defaults to `$(PREFIX)/share` and `$(DATADIR)/doc/samurai` respectively as per the GNU install standards.

https://www.gnu.org/software/make/manual/html_node/Directory-Variables.html

Adding `DATAROOTDIR` is probably not needed since samurai does not ship any project specific data files.

Note: Any distros or other users that ship samurai without including the license file are technically committing a license violation of all four included licenses.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,9 @@
 
 PREFIX=/usr/local
 BINDIR=$(PREFIX)/bin
-MANDIR=$(PREFIX)/share/man
+DATADIR=$(PREFIX)/share
+DOCDIR=$(DATADIR)/doc/samurai
+MANDIR=$(DATADIR)/man
 ALL_CFLAGS=$(CFLAGS) -std=c99 -Wall -Wextra -Wpedantic -Wno-unused-parameter
 OBJ=\
 	build.o\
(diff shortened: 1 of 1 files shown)


michaelforney:
Thanks for the PR, but it does not seem to be common practice for the package itself to install the license file, rather this is something that distributions take care of. I looked at several other notable packages using this style license (ninja, curl, ncurses, openssh, libxkbcommon) and none of them install their license themselves.

orbea:
Yea, it is not a common practice, implementing `docdir` at all is very hit and miss. Regardless some examples I found that do install the license by default I found after a brief search are `sudo`, `libjpeg-turbo` and `libsamplerate`.

> rather this is something that distributions take care of.

Often this is the case, but in my experience I have gotten some very hostile responses when bringing this up hence what inspired this PR...

eli-schwartz:
On Arch Linux, missing licenses are a lint warning unless they're installed to `/usr/share/licenses/<name-of-package>`. Is docdir common for installing licenses on other distros?

orbea:
>  Is docdir common for installing licenses on other distros?

When licenses are installed by the `install` target the install path is usually controlled with some variation of `--docdir`. However many projects do not install licenses by default nor utilize `--docdir` at all, meson even does not have `--docdir` which is rather unfortunate...

Personally I prefer when `--docdir` is available and used, but I don't speak for everyone obviously.

eli-schwartz:
I did not ask what the name of the flag is, I asked about the philosophical concept of a "documentation directory"...

So you'd recommend I pass an option to install the project README (and also the license file) to /usr/share/licenses/samurai/README.md by defining it as the "directory for installing software documentation"?

orbea:
That is a good point which I didn't not consider before making this PR. In practice I have only seen `--docdir` used, but this was discussed for the sofort build system (https://dev.midipix.org/zippy/sofort/blob/main/f/README.SOFORT) that something like `--licensedir` could be also added which then could default to `$docdir` to keep the historical behavior while still allowing the behavior distros like arch expect which installs the license to a different directory than other documentation. I'm not sure anything was done there yet regarding this.

PR #78, opened 2021-10-04
manual: Use Pq macro to parenthesize link.

averms:
This fixes a visual bug in how the man page renders on my Linux machine (which uses groff(1) and not mandoc(1)). This is how it looks in master:
image

Not sure if this is a groff bug but the patch makes it work with both mandoc(1) and groff(1)


michaelforney:
Thanks for the patch, merged.

PR #80, opened 2021-11-10
Implement graph tool

nico-abram:
This pull request should implement the graph tool

### DIFF
--- a/tool.c
+++ b/tool.c
@@ -172,6 +172,72 @@ commands(int argc, char *argv[])
 }
 
 static void
+graphviznode(struct node *n)
+{
+	size_t i;
+	char *order_only;
+	struct edge *e;
+
+	printf("\"%p\" [label=\"%s\"]\n", (void *)n, n->path->s);
+
+	e = n->gen;
+	if (!e || (e->flags & FLAG_WORK))
+		return;
+	e->flags |= FLAG_WORK;
+
+	for (i = 0; i < e->nin; i++)
+		graphviznode(e->in[i]);
+
+	if (e->nin == 1 && e->nout == 1) {
+		printf("\"%p\" -> \"%p\" [label=\" %s\"]\n", (void *)e->in[0], (void *)e->out[0], e->rule->name);
+	} else {
+		printf("\"%p\" [label=\"%s\", shape=ellipse]\n", (void *)e, e->rule->name);
+		for (i = 0; i < e->nout; i++)
+			printf("\"%p\" -> \"%p\"\n", (void *)e, (void *)e->out[i]);
+		for (i = 0; i < e->nin; i++) {
+			order_only = "";
+			if (i >= e->inorderidx)
+				order_only = " style=dotted";
+			printf("\"%p\" -> \"%p\" [arrowhead=none%s]\n", (void *)e->in[i], (void *)e, order_only);
+		}
+	}
+}
+
+static int
+graphviz(int argc, char *argv[])
+{
+	int ret = 0;
+	struct node *n;
+
+	puts("digraph ninja {");
+	puts("rankdir=\"LR\"");
+	puts("node [fontsize=10, shape=box, height=0.25]");
+	puts("edge [fontsize=10]");
+
+	if (argc) {
+		while (*++argv) {
+			n = nodeget(*argv, 0);
+			if (!n) {
+				warn("unknown target '%s'", *argv);
+				ret = 1;
+				continue;
+			}
+			graphviznode(n);
+		}
+	} else {
+		defaultnodes(graphviznode);
+	}
+
+	puts("}");
+
+	fflush(stdout);
+	if (ferror(stdout))
+		fatal("write failed");
+
+	return ret;
+}
+
+static void
 printjson(const char *s, size_t n, bool join)
 {
 	size_t i;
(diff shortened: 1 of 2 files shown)


tool.c
```diff
@@ -171,6 +171,72 @@ commands(int argc, char *argv[])
 	return ret;
 }
 
+static void
+graphviznode(struct node *n)
+{
+	size_t i;
+	char *order_only;
```
michaelforney, line 178: I'd make this `const char *`, and maybe call it `style`.

tool.c
```diff
@@ -171,6 +171,72 @@ commands(int argc, char *argv[])
...
+	if (e->nin == 1 && e->nout == 1) {
+		printf("\"%p\" -> \"%p\" [label=\" %s\"]\n", (void *)e->in[0], (void *)e->out[0], e->rule->name);
+	} else {
+		printf("\"%p\" [label=\"%s\", shape=ellipse]\n", (void *)e, e->rule->name);
+		for (i = 0; i < e->nout; i++)
+			printf("\"%p\" -> \"%p\"\n", (void *)e, (void *)e->out[i]);
+		for (i = 0; i < e->nin; i++) {
+			order_only = "";
+			if (i >= e->inorderidx)
+				order_only = " style=dotted";
```
michaelforney, line 200: This seems like a good place to use the conditional operator:
```
order_only = i >= e->inorderidx ? " style=dotted" : "";
```

tool.c
```diff
@@ -171,6 +171,72 @@ commands(int argc, char *argv[])
...
+			graphviznode(n);
+		}
+	} else {
+		defaultnodes(graphviznode);
+	}
+
+	puts("}");
+
+	fflush(stdout);
+	if (ferror(stdout))
```
michaelforney, line 234: You could combine these lines to match the rest of the tools:
```
if (fflush(stdout) || ferror(stdout))
```

tool.c
```diff
@@ -171,6 +171,72 @@ commands(int argc, char *argv[])
...
+	struct edge *e;
+
+	printf("\"%p\" [label=\"%s\"]\n", (void *)n, n->path->s);
+
+	e = n->gen;
+	if (!e || (e->flags & FLAG_WORK))
+		return;
+	e->flags |= FLAG_WORK;
+
+	for (i = 0; i < e->nin; i++)
```
michaelforney, line 188: Minor nitpick: use `++i` for consistency with the rest of the codebase.

tool.c
```diff
@@ -171,6 +171,72 @@ commands(int argc, char *argv[])
...
+{
+	int ret = 0;
+	struct node *n;
+
+	puts("digraph ninja {");
+	puts("rankdir=\"LR\"");
+	puts("node [fontsize=10, shape=box, height=0.25]");
+	puts("edge [fontsize=10]");
+
+	if (argc) {
```
michaelforney, line 217: This needs to be `argc > 1`, since `argv[0]` is always the tool name.

tool.c
```diff
@@ -171,6 +171,72 @@ commands(int argc, char *argv[])
...
+	e = n->gen;
+	if (!e || (e->flags & FLAG_WORK))
+		return;
+	e->flags |= FLAG_WORK;
+
+	for (i = 0; i < e->nin; i++)
+		graphviznode(e->in[i]);
+
+	if (e->nin == 1 && e->nout == 1) {
+		printf("\"%p\" -> \"%p\" [label=\" %s\"]\n", (void *)e->in[0], (void *)e->out[0], e->rule->name);
```
michaelforney, line 192: Do you know the reason for the leading space? Ninja says it's for aesthetics, but it seems to look fine to me without it.

tool.c
```diff
@@ -171,6 +171,72 @@ commands(int argc, char *argv[])
...
+			order_only = "";
+			if (i >= e->inorderidx)
+				order_only = " style=dotted";
+			printf("\"%p\" -> \"%p\" [arrowhead=none%s]\n", (void *)e->in[i], (void *)e, order_only);
+		}
+	}
+}
+
+static int
+graphviz(int argc, char *argv[])
```
michaelforney, line 207: I'd call this `graph` (even though it is a bit generic), so that the function name always matches the tool name.

tool.c
```diff
@@ -387,6 +453,7 @@ static const struct tool tools[] = {
 	{"compdb", "dump compilation database", compdb},
 	{"query", "show incoming/outgoing edges for a path", query},
 	{"targets", "list targets", targets},
+	{"graph", "output graphviz dot file for targets", graphviz},
```
michaelforney, line 456: Could you sort this by tool name? Also, try to order the functions in the same order they appear in this list. I know `query` and `targets` are currently out of order, but I'll fix this after merging this PR.

michaelforney (commented):
This looks pretty good, thanks for working on it!

I have a few minor stylistic comments, but apart from those, we just need to document this tool in the manual and this should be good to go.

tool.c
nico-abram: I don't, I was following ninja, looks fine without to me too. Removed it for now.

tool.c
nico-abram: Also renamed the recursive node function

nico-abram:
Thanks for the review. Pushed a separate commit with the fixes to ease review (Although since I moved the functions I don't think it helps). I can squash if needed.

Also added the manual entry, apologies if I did it wrong, I've never touched man pages before.

michaelforney:
Looks good, merged. Thanks again!

Issue #41, opened 2020-02-13
Implement `-l` option (job scheduling based on load average)


orbea:
Is there any reason samu does not support the `-l` argument as ninja does?

From `ninja --help`.
> -l N     do not start new jobs if the load average is greater than N

michaelforney:
No reason other than I didn't implement it.

It would have to be guarded by an ifdef since getloadavg is non-POSIX.

orbea:
Thanks for the reply, I did not realize that about getloadavg. I'm fine with whatever you decide is best here, but I was curious as I was told this is part of the gentoo eclass for ninja and then noticed that samurai does not support it.

orbea:
@michaelforney Maybe it might be nice for at least the short term it would be implemented as a no-op and prints a short warning about not being implemented instead of outright failing? This might help with compatiblity and ninja.

Salamandar:
I'm wondering if I'd have some time to implement it. The answer is probably yes. That wouldn't be very difficult.

Salamandar:
I'm working in #50.

PR #82, opened 2022-02-21
Load average option

capezotte:
This is an alternative implementation of #41, and owes quite a bit of code to #50.

This avoids having multiple threads and instead relies on making `poll` time out.

I'm not sure of the best way to guard the load-average-using code, though. There doesn't seem to be a macro associated with it (and forgive me if my lack of experience with C is showing). That code is still unguarded in this PR.

### DIFF
--- a/build.c
+++ b/build.c
@@ -1,4 +1,7 @@
 #define _POSIX_C_SOURCE 200809L
+#ifndef NO_GETLOADAVG
+#define _BSD_SOURCE /* for getloadavg */
+#endif
 #include <errno.h>
 #include <fcntl.h>
 #include <inttypes.h>
@@ -520,12 +523,28 @@ done:
 	return false;
 }
 
+#ifndef NO_GETLOADAVG
+/* Queries the system load average. */
+static double
+queryload(void)
+{
+	double load;
+
+	if (getloadavg(&load, 1) == -1) {
+		warn("getloadavg:");
+		load = 100.0;
+	}
+
+	return load;
+}
+#endif
+
 void
 build(void)
 {
 	struct job *jobs = NULL;
 	struct pollfd *fds = NULL;
-	size_t i, next = 0, jobslen = 0, numjobs = 0, numfail = 0;
+	size_t i, next = 0, jobslen = 0, maxjobs = buildopts.maxjobs, numjobs = 0, numfail = 0;
 	struct edge *e;
 
 	if (ntotal == 0) {
@@ -538,8 +557,13 @@ build(void)
 
 	nstarted = 0;
 	for (;;) {
+#ifndef NO_GETLOADAVG
+		/* limit number of of jobs based on load */
+		if (buildopts.maxload)
+			maxjobs = queryload() > buildopts.maxload ? 1 : buildopts.maxjobs;
+#endif
 		/* start ready edges */
-		while (work && numjobs < buildopts.maxjobs && numfail < buildopts.maxfail) {
+		while (work && numjobs < maxjobs && numfail < buildopts.maxfail) {
 			e = work;
 			work = work->worknext;
 			if (e->rule != &phonyrule && buildopts.dryrun) {
@@ -578,7 +602,7 @@ build(void)
 		}
 		if (numjobs == 0)
 			break;
-		if (poll(fds, jobslen, -1) < 0)
+		if (poll(fds, jobslen, 5000) < 0)
 			fatal("poll:");
 		for (i = 0; i < jobslen; ++i) {
 			if (!fds[i].revents || jobwork(&jobs[i]))
--- a/samu.c
+++ b/samu.c
@@ -20,7 +20,7 @@ const char *argv0;
 static void
 usage(void)
 {
-	fprintf(stderr, "usage: %s [-C dir] [-f buildfile] [-j maxjobs] [-k maxfail] [-n]\n", argv0);
+	fprintf(stderr, "usage: %s [-C dir] [-f buildfile] [-j maxjobs] [-k maxfail] [-l maxload] [-n]\n", argv0);
 	exit(2);
 }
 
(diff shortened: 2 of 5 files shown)


samu.c
```diff
@@ -163,8 +179,7 @@ main(int argc, char *argv[])
 		buildopts.maxfail = num > 0 ? num : -1;
 		break;
 	case 'l':
-		warn("job scheduling based on load average is not implemented");
-		EARGF(usage());
+		loadflag(EARGF(usage()));
```
michaelforney, line 182: I'd inline the body of `loadflag` here, to match `-k`.

samu.c
```diff
@@ -20,7 +20,7 @@ const char *argv0;
 static void
 usage(void)
 {
-	fprintf(stderr, "usage: %s [-C dir] [-f buildfile] [-j maxjobs] [-k maxfail] [-n]\n", argv0);
+	fprintf(stderr, "usage: %s [-C dir] [-f buildfile] [-j maxjobs] [-l maxload] [-k maxfail] [-n]\n", argv0);
```
michaelforney, line 23: Maybe move this after the `-k` flag?

build.c
```diff
@@ -520,6 +522,25 @@ jobwork(struct job *j)
 	return false;
 }
 
+/* Queries the system load average. */
+static void
+queryload()
```
michaelforney, line 527: Should use a prototype declaration here (`static void queryload(void)`).

build.c
```diff
@@ -575,11 +596,13 @@ build(void)
 				next = jobs[next].next;
 				++numjobs;
 			}
+
```
michaelforney, line 599: Extra newline snuck in here.

build.c
```diff
@@ -28,11 +29,12 @@ struct job {
 	bool failed;
 };
 
-struct buildoptions buildopts = {.maxfail = 1};
+struct buildoptions buildopts = {.maxfail = 1, .maxload = 0.0 };
```
michaelforney, line 32: 0 is the default value for `maxload`, so no need to specify it explicitly here.

build.c
```diff
@@ -539,7 +560,7 @@ build(void)
 	nstarted = 0;
 	for (;;) {
 		/* start ready edges */
-		while (work && numjobs < buildopts.maxjobs && numfail < buildopts.maxfail) {
+		while (work && (numjobs == 0 || loadavglow) && numjobs < buildopts.maxjobs && numfail < buildopts.maxfail) {
```
michaelforney, line 563: Rather than this `loadvglow` global, I think I'd prefer if we just had the `queryload` function return the load as a double. Then at the start of this loop, do something like

```
maxjobs = buildopts.maxload && queryload() > buildopts.maxload ? 1 : buildopts.maxjobs;
```

and use `maxjobs` instead of `buildopts.maxjobs` in the while loop. What do you think of that approach?

michaelforney (commented):
Thanks, this looks really good. Much simpler than the other PR.

samu.c
capezotte: Should it also be inside the body of `parseenvargs()` (or be dropped from SAMUFLAGS)?

build.c
capezotte: Makes sense, we can even mimic Ninja's load average logic and use something like:

     (numjobs == 0 || buildopts.maxload <= 0.0 || queryload() < buildopts.maxload)

samu.c
eli-schwartz: Since the load average option is effectively the same as the jobs option (in the sense that both have no effect on the build output, only on job scheduling, parallelization) it makes a lot of sense to support -l in the SAMUFLAGS as well.

So yes, I believe it should be read in `parseenvargs()`.

I think there's probably not a good reason to ever drop anything from SAMUFLAGS -- I assume by that you mean silently permit its presence but ignore it.

michaelforney: Yep, I just forgot about `SAMUFLAGS`. It's fine as it is.

build.c
michaelforney: I think both logic works out the same, right? We start a job if there are none currently running, or if we are under the max load and max jobs.

The reason I suggested the former is to avoid the really long condition.

capezotte:
Once we get these initial issues sorted out, I'd like to see what options would be best for  for guarding this code.

capezotte:
I'm thinking of guarding this commit's additions with `#ifdef HAS_GETLOADAVG` and documenting the need for adding `-DHAS_GETLOADAVG` somewhere. Does this sound acceptable to you?

build.c
capezotte: Changed to add a `maxjobs` variable and an `if` statement. Going to mark this as resolved, but feel free to comment on the newer version if you disagree.

michaelforney:
> I'm thinking of guarding this commit's additions with `#ifdef HAS_GETLOADAVG` and documenting the need for adding `-DHAS_GETLOADAVG` somewhere. Does this sound acceptable to you?

Since this API is pretty widely available, I think we could do the opposite and introduce a `NO_GETLOADAVG` to disable its use.

(sorry for the late reply, I've been a bit busy lately)

capezotte:
Guards are in place now.

capezotte:
I'm satisfied with the state of this PR.

orbea:
@michaelforney What's the status of this PR?

michaelforney:
@capezotte Any idea if macOS has `getloadavg`, and if so, how to expose it?

Currently it is failing due to missing declaration: https://github.com/michaelforney/samurai/runs/7437461267

I'm not quite sure if we should set `-DNO_GETLOADAVG` for the mac build, or if some other define or include is needed for it.

capezotte:
I looked into [Apple's stdlib.h](https://opensource.apple.com/source/Libc/Libc-1439.40.11/include/stdlib.h.auto.html) and it seems to want us to `#define _DARWIN_C_SOURCE` to get access to `getloadavg`.

By the way, I checked the Ubuntu build and glibc prints a warning for defining `_BSD_SOURCE`.

So it should look like this:

```c
#ifndef NO_GETLOADAVG
#	if defined(__APPLE__) && defined(__MACH__)
#		define _DARWIN_C_SOURCE
#	else
#		define _BSD_SOURCE
#		define _DEFAULT_SOURCE /* silence glibc >= 2.19 warning */
#	endif
#endif
```

PR #87, opened 2022-07-13
build: Update warning flags

orbea:
This has 3 changes.

~~1. Remove `-Wno-unused-parameter`. There are currently no warnings and its probably better to solve these with `if (foo) {}` in the cases they exist to prevent hiding legitimate warnings.~~
2. Adds `-Wshadow` and fixes the one warning revealed by both gcc and clang.
3. Adds `-Wmissing-prototypes` which doesn't reveal any existing warnings, but it would be nice to not inadvertently introduce it in the future.
4. Fixes a `-Wcpp` warning.

This was tested on Slackware with gcc-11.2.0/clang-13.0.0 and Gentoo with gcc-11.3.0/clang-14.0.6.

The fixed `-Wshadow` warning:
```
samu.c: In function ‘main’:
samu.c:187:21: warning: declaration of ‘n’ shadows a previous local [-Wshadow]
  187 |                 int n = sysconf(_SC_NPROCESSORS_ONLN);
      |                     ^
samu.c:126:22: note: shadowed declaration is here
  126 |         struct node *n;
      |                      ^
```
```
In file included from /usr/include/errno.h:25,
                 from build.c:5:
/usr/include/features.h:187:3: warning: #warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE" [-Wcpp]
  187 | # warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE"
      |   ^~~~~~~
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 PREFIX=/usr/local
 BINDIR=$(PREFIX)/bin
 MANDIR=$(PREFIX)/share/man
-ALL_CFLAGS=$(CFLAGS) -std=c99 -Wall -Wextra -Wpedantic -Wno-unused-parameter
+ALL_CFLAGS=$(CFLAGS) -std=c99 -Wall -Wextra -Wshadow -Wpedantic -Wno-unused-parameter
 OBJ=\
 	build.o\
 	deps.o\
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 PREFIX=/usr/local
 BINDIR=$(PREFIX)/bin
 MANDIR=$(PREFIX)/share/man
-ALL_CFLAGS=$(CFLAGS) -std=c99 -Wall -Wextra -Wshadow -Wpedantic -Wno-unused-parameter
+ALL_CFLAGS=$(CFLAGS) -std=c99 -Wall -Wextra -Wshadow -Wmissing-prototypes -Wpedantic -Wno-unused-parameter
 OBJ=\
 	build.o\
 	deps.o\
--- a/build.c
+++ b/build.c
@@ -1,6 +1,12 @@
 #define _POSIX_C_SOURCE 200809L
+/* expose getloadavg */
 #ifndef NO_GETLOADAVG
-#define _BSD_SOURCE /* for getloadavg */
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -4,7 +4,7 @@
 PREFIX=/usr/local
 BINDIR=$(PREFIX)/bin
 MANDIR=$(PREFIX)/share/man
-ALL_CFLAGS=$(CFLAGS) -std=c99 -Wall -Wextra -Wshadow -Wpedantic
+ALL_CFLAGS=-std=c99 -Wall -Wextra -Wshadow -Wmissing-prototypes -Wpedantic
```
michaelforney, line 7: Why remove $(CFLAGS) from here? `ALL_CFLAGS` is intended to contain all flags needed to compile the source files.

Makefile
```diff
@@ -4,7 +4,7 @@
 PREFIX=/usr/local
 BINDIR=$(PREFIX)/bin
 MANDIR=$(PREFIX)/share/man
-ALL_CFLAGS=$(CFLAGS) -std=c99 -Wall -Wextra -Wpedantic -Wno-unused-parameter
```
michaelforney, line 7: I just pushed the load average commit, which will now cause a warning when you build with `-DNO_GETLOADAVG`.

I disagree that unused parameters are something that should be warned about. In my opinion, working around the issue with no-op code or comments to silence the warning is worse than just disabling them.

michaelforney (commented):
I'm happy to add `-Wshadow` and `-Wmissing-prototypes`, but I think `-Wno-unused-parameter` should stay.

Makefile
orbea: I moved it down to the actual rule so that the line would be shorter, but I moved it back now.

Makefile
orbea: I added a commit to fix the warning, I think this is the right way to do it?

orbea:
> but I think -Wno-unused-parameter should stay.

I added it back.

michaelforney:
Thanks! Merged.

Issue #85, opened 2022-06-15
Windows: msys2 + mingw-w64: warning


vtorri:
I don't know if you want a warning free compilation, but here it is:

c99 -O1 -std=c99 -Wall -Wextra -Wpedantic -Wno-unused-parameter -c -o graph.o graph.c
In file included from graph.c:2:
graph.c: In function ‘nodepath’:
graph.c:122:30: warning: array subscript has type ‘char’ [-Wchar-subscripts]
  122 |                 if (!isalnum(*s) && !strchr("_+-./", *s))
      |                              ^~

just cast *s to int

vtorri:
i can do a PR if you want

michaelforney:
Yes, a PR would be appreciated. Thanks!

vtorri:
@michaelforney see #88

PR #88, opened 2022-07-21
Fix warning, see issue #85

vtorri:


vtorri:
any news on this simple PR ?

graph.c
```diff
@@ -119,7 +119,7 @@ nodepath(struct node *n, bool escape)
 	escape = false;
 	nquote = 0;
 	for (s = n->path->s; *s; ++s) {
-		if (!isalnum(*s) && !strchr("_+-./", *s))
+		if (!isalnum((int)(*s)) && !strchr("_+-./", *s))
```
E5ten, line 122: The parens around `*s` aren't necessary

vtorri: and ?

E5ten: And they should be removed, because redundant parens next to casts aren't used elsewhere in the code.

vtorri: let's wait for @michaelforney 's answer

michaelforney, line 122: The warning is legitimate, but this is not a correct fix. The C standard says that if the value is not representable as `unsigned char`, we get undefined behavior. If `char` is signed, and has a negative value (for example, one of the code units of some multibyte UTF-8 sequence), then converting to `int` still retains the negative value.

Instead, we want to interpret the the pointed-to `char` as `unsigned char`:

```C
isalnum(*(unsigned char *)s);
```

michaelforney: Yes, please remove the unnecessary parentheses to be consistent with the rest of the code.

vtorri:
@michaelforney should be good now

michaelforney:
Thanks! Applied.

PR #93, opened 2022-10-03
Add -q option to supress output (like ninja --quiet)

annacrombie:


michaelforney:
Another way to get this functionality is `>/dev/null`.

Issue #71, opened 2021-04-14
Support make jobserver protocol


rossburton:
GNU Make has a neat feature called the jobserver protocol, where the top-level Make can allocate a specific number of job slots, and child makes can take slots to do work in. This was designed to stop the parallelisation problem where a top-level `make -j10` may potentially spawn 10 separate sub-makes all with `-j10` so there's now 100 parallel jobs.

However, it's also useful for resource control is systems which built multiple pieces of software at once.  For example, Bitbake can build N different pieces of software at once, and each of those is passed a `-jM` flag. If each of these N tasks is compiling then thats's N*M jobs so you don't want N or M to be too high, but if only 1 of N is building then you want M to be high.

With the job server protocol there are N slots in total for all sub makes, so you can control the resource utilisation more accurately.  If Samurai also supported the jobserver protocol instead of just `-j` then it could join in the resource pooling and builds can be more efficient.

The canonical implementation of jobserver is in the GNU Make sources. The documentation is pretty vague as it is considered an implementation detail, but it hasn't changed in 20 years...  The prototype BitBake support for a master jobserver queue is at http://git.yoctoproject.org/cgit.cgi/poky-contrib/commit/?h=rpurdie/wipqueue4&id=d66a327fb6189db5de8bc489859235dcba306237

rossburton:
cc @rpurdie

bonzini:
Hi, I have done a quick and dirty implementation (less than 200 lines of code) but it uses pthreads. Should I submit a pull request or should I first try to rework it without threads?

michaelforney:
@bonzini, I am interested in taking a look at what you've got, but I'd definitely prefer an implementation without threads.

eli-schwartz:
BTW since https://git.savannah.gnu.org/cgit/make.git/commit/?id=7ad2593b2d2bb5b9332f4444d8bf93ac6f958bc6 GNU make supports named pipes as an alternative to anonymous pipes.

Once it actually gets into a release that should be pretty nifty.

mid-kid:
I feel it's worth noting that gcc picks up on this jobserver when using `gcc -flto` for LTO builds, and as such is very useful for speeding that up. Using `-flto=n` to specify the amount of LTO linking jobs isn't great when it comes to projects linking multiple programs concurrently.

bonzini:
> @bonzini, I am interested in taking a look at what you've got, but I'd definitely prefer an implementation without threads.

[...]

PR #94, opened 2022-10-28
Support for GNU Make jobserver

capezotte:
Quick and dirty implementation of #71, following the GNU Make 4.3 manual (Section 13.1.

TODO:
- [ ] `-l` and `-j` are ignored when jobserver options are active, though CMake team's proposed patch for Ninja instead ignores jobserver options whenever -j is given in the command line. Should we follow them?
- [x] Aborts whenever given jobserver FDs are closed/invalid. Should it be downgraded to warning?
- [x] Make sure it works with older GNU Make (pre-4.2, with different option names).
- [x] The `fifo:` protocol mentioned on #71.
- [x] Tokens should be returned to GNU Make when the program terminates. This is probably the biggest issue with this PR, but I'd like feedback on the "core" of the changes before embarking on rewriting `fatal` and installing signal handlers.

### DIFF
--- a/build.c
+++ b/build.c
@@ -549,23 +549,28 @@ queryload(void)
 }
 
 /* returns remaining GNU/tokens */
-static void
+static ssize_t
 returngmake(void)
 {
-	if (buildopts.gmakepipe[1] >= 0) {
-		if (write(buildopts.gmakepipe[1], gmaketokens, gmakelatest - gmaketokens) < 0) {
-			warn("last write to jobserver:");
-		} else {
+	ssize_t returned = 0;
+	if (buildopts.gmakepipe[1] >= 0)
+		if ((returned = write(buildopts.gmakepipe[1], gmaketokens, gmakelatest - gmaketokens)) >= 0)
 			gmakelatest = gmaketokens;
-		}
-	}
+	return returned;
+}
+
+static void
+gmakeatexit(void)
+{
+	if (returngmake() < 0)
+		warn("last write to jobserver:");
 }
 
 static void
 termsignal(int signum)
 {
-	write(2, "samu: terminating due to signal\n", 32);
-	returngmake();
+	write(2, "terminating due to signal\n", 27);
+	(void)returngmake();
 	_exit(128 + signum);
 }
 
@@ -585,7 +590,7 @@ build(void)
 	}
 
 	if (buildopts.gmakepipe[0] >= 0) {
-		if (atexit(returngmake) == 0) {
+		if (atexit(gmakeatexit) == 0) {
 			maxjobs = 1; /* will change dynamically as tokens are exchanged */
 			tokenin.fd = buildopts.gmakepipe[0];
 			sact.sa_handler = termsignal;
--- a/build.c
+++ b/build.c
@@ -327,15 +327,6 @@ jobstart(struct job *j, struct edge *e)
 		warn("posix_spawn_file_actions_addclose:");
 		goto err3;
 	}
-	if (isjobserverclient()) {
-		/* do not allow children to steal GNU/tokens */
-		for (i = 0; i < 2; ++i) {
-			if ((errno = posix_spawn_file_actions_addclose(&actions, buildopts.gmakepipe[i]))) {
-				warn("posix_spawn_file_actions_addclose:");
-				goto err3;
-			}
-		}
-	}
 	if (e->pool != &consolepool) {
 		if ((errno = posix_spawn_file_actions_addopen(&actions, 0, "/dev/null", O_RDONLY, 0))) {
 			warn("posix_spawn_file_actions_addopen:");
@@ -538,7 +529,7 @@ done:
 }
 
 /* queries the system load average */
-static double
+static bool
 reachedload(void)
 {
 #ifdef HAVE_GETLOADAVG
@@ -603,7 +594,7 @@ build(void)
 	fds[0].fd = buildopts.gmakepipe[0];
 	fds[0].events = 0;
 	if (isjobserverclient()) {
-		fds[0].events &= POLLIN;
+		fds[0].events |= POLLIN;
 		if (atexit(gmakeatexit) == 0) {
 			sact.sa_handler = termsignal;
 			sigaction(SIGTERM, &sact, NULL);
--- a/build.c
+++ b/build.c
@@ -36,6 +36,7 @@ static struct edge *work;
 static size_t nstarted, nfinished, ntotal;
 static bool consoleused;
 static struct timespec starttime;
+static char gmaketokens[512], *gmakelatest = gmaketokens;
 
 void
 buildreset(void)
@@ -322,6 +323,15 @@ jobstart(struct job *j, struct edge *e)
 		warn("posix_spawn_file_actions_addclose:");
 		goto err3;
 	}
+	if (buildopts.gmakepipe[0] >= 0) {
+		/* do not allow children to steal GNU/tokens */
+		for (i = 0; i < 2; ++i) {
+			if ((errno = posix_spawn_file_actions_addclose(&actions, buildopts.gmakepipe[i]))) {
+				warn("posix_spawn_file_actions_addclose:");
+				goto err3;
+			}
+		}
+	}
 	if (e->pool != &consolepool) {
 		if ((errno = posix_spawn_file_actions_addopen(&actions, 0, "/dev/null", O_RDONLY, 0))) {
 			warn("posix_spawn_file_actions_addopen:");
@@ -541,14 +551,24 @@ queryload(void)
 #endif
 }
 
+/* returns remaining GNU/tokens */
+static void
+returngmake(void)
+{
+	if (buildopts.gmakepipe[1] >= 0) {
+		if (write(buildopts.gmakepipe[1], gmaketokens, gmakelatest - gmaketokens) < 0)
+			warn("last write to jobserver:");
+	}
+}
+
 void
 build(void)
 {
 	struct job *jobs = NULL;
 	struct pollfd *fds = NULL, tokenin = { .fd = -1, .events = POLLIN };
-	size_t i, next = 0, jobslen = 0, maxjobs = buildopts.maxjobs, numjobs = 0, numfail = 0, gmakeread;
+	size_t i, next = 0, jobslen = 0, maxjobs = buildopts.maxjobs, numjobs = 0, numfail = 0;
+	ssize_t gmakeread;
 	struct edge *e;
-	char gmaketokens[512], *gmakelatest = gmaketokens;
 
 	if (ntotal == 0) {
 		warn("nothing to do");
@@ -556,8 +576,12 @@ build(void)
 	}
 
 	if (buildopts.gmakepipe[0] >= 0) {
-		maxjobs = 1; /* will change dynamically as tokens are exchanged */
-		tokenin.fd = buildopts.gmakepipe[0];
+		if (atexit(returngmake) == 0) {
+			maxjobs = 1; /* will change dynamically as tokens are exchanged */
+			tokenin.fd = buildopts.gmakepipe[0];
+		} else {
+			warn("unable to register an atexit() function");
+		}
 	}
 
 	clock_gettime(CLOCK_MONOTONIC, &starttime);
@@ -568,7 +592,7 @@ build(void)
 		/* see if GNU/make has authorized more jobs */
 		if (poll(&tokenin, 1, 0) > 0) {
 			gmakeread = gmakelatest - gmaketokens;
-			gmakeread = read(buildopts.gmakepipe[0], gmakelatest, 512 - gmakeread > ntotal - nstarted ? ntotal - nstarted : 512 - gmakeread);
+			gmakeread = read(buildopts.gmakepipe[0], gmakelatest, 512 - (size_t)gmakeread > ntotal - nstarted ? ntotal - nstarted : 512 - gmakeread);
 			if (gmakeread < 0) {
 				warn("read from jobserver:");
 				gmakeread = 0;
@@ -637,12 +661,6 @@ build(void)
 				}
 				--maxjobs;
 			}
-		}
-	}
-	/* return remaining GNU/tokens */
-	if (buildopts.gmakepipe[1] >= 0) {
-		if (write(buildopts.gmakepipe[1], gmaketokens, gmakelatest - gmaketokens) < 0) {
-			warn("last write to jobserver:");
 		}
 	}
 	for (i = 0; i < jobslen; ++i)
--- a/build.c
+++ b/build.c
@@ -556,14 +556,26 @@ static void
 returngmake(void)
 {
 	if (buildopts.gmakepipe[1] >= 0) {
-		if (write(buildopts.gmakepipe[1], gmaketokens, gmakelatest - gmaketokens) < 0)
+		if (write(buildopts.gmakepipe[1], gmaketokens, gmakelatest - gmaketokens) < 0) {
 			warn("last write to jobserver:");
+		} else {
+			gmakelatest = gmaketokens;
+		}
 	}
+}
+
+static void
+termsignal(int signum)
+{
+	write(2, "samu: terminating due to signal\n", 32);
+	returngmake();
+	_exit(128 + signum);
 }
 
 void
 build(void)
 {
+	struct sigaction sact = { 0 };
 	struct job *jobs = NULL;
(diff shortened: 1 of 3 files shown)


michaelforney:
This looks really good, thanks for working on it.

> Some still-relevant distros ship with older GNU Make (pre-4.2, with different option names).

Can you elaborate? What happens if `samu` gets run with an old GNU make? Does it break, or just not work?

> Tokens aren't returned to GNU Make when the program terminates.

Can we install an `atexit` function to clean them up?

capezotte:
> Can you elaborate? What happens if samu gets run with an old GNU make? Does it break, or just not work?

Old GNU make uses `--jobserver-fds=` instead of `--jobserver-auth=` in the environment. The protocol isn't documented in the old version of manual, so I was unsure.

After some testing and a cursory reading of the source code, it seems that, other than that, it's the same.

> Can we install an atexit function to clean them up?

Good idea, just implemented it and tokens are returned after `fatal` and after a normal exit.

build.c
```diff
@@ -541,24 +551,71 @@ queryload(void)
 #endif
 }
 
+/* returns remaining GNU/tokens */
+static void
+returngmake(void)
+{
+	if (buildopts.gmakepipe[1] >= 0) {
+		if (write(buildopts.gmakepipe[1], gmaketokens, gmakelatest - gmaketokens) < 0) {
+			warn("last write to jobserver:");
```
nekopsykose, line 560: this is called in a signal handler (termsignal, below), but warn() then uses the printf(3) family, which is not signal-safe. isn't this technically illegal? (it exits immediately after, so maybe that's ok?)

thesamesam:
@michaelforney are you able to take a look? thank you!

michaelforney:
@thesamesam There are several outstanding comments that still need to be resolved.

build.c
capezotte: I've moved the warn() part to a separate function.

trws:
What remains to be done here? Sorry for the drive-by comment, but I've been looking for a ninja alternative with this feature to use with our package manager, which allows doing a build of an entire tree of packages under a single jobserver.  If I could help out a bit, I'd be happy to.

One note, this line: https://github.com/michaelforney/samurai/pull/94/files#diff-99b168f07bbb2d0e68bfe3cd8378543a6becd5989922c5283e499d1f69879b42R324
mentions children "stealing" tokens.  If possible, it would be best to allow children to take tokens.  This implementation allows them to with the gnu make 4.3 style interface, but not the old one.  Reason being that things like cargo, spack, make, ld.gold and others can use that for their own parallelism instead of doing their own fanout and over-loading the system.  

On a related note, would you like to have jobserver server support here as well?

michaelforney:
> What remains to be done here? Sorry for the drive-by comment, but I've been looking for a ninja alternative with this feature to use with our package manager, which allows doing a build of an entire tree of packages under a single jobserver. If I could help out a bit, I'd be happy to.

See the unresolved comments. In particular, the one about polling. There should be a single poll in the event loop.

I pushed my current working tree to the jobserver branch. I don't remember the state, but if someone else wants to help out with this, that'd be a good place to start.

michaelforney (commented):
Well, turns out that the comments I thought I made were never submitted. My apologies.

michaelforney:
One question I have, should a jobserver client request a token for every job it starts and release it when its done? Or is it fine if samurai takes as many jobs as it can use, and release them only when it has extra tokens that it can't use?

I don't think it would make a difference, except maybe collective ordering of the jobs run.

eli-schwartz:
If it's going to immediately reuse those tokens then I think it's fine.

build.c
```diff
@@ -610,6 +672,13 @@ build(void)
 			next = i;
 			if (jobs[i].failed)
 				++numfail;
+			/* must return the GNU/tokens once a job is done */
+			if (buildopts.gmakepipe[1] > 0 && gmakelatest != gmaketokens) {
+				if (write(buildopts.gmakepipe[1], --gmakelatest, 1) < 0) {
```
michaelforney, line 677: I'm not sure if it matters, but this might not be the same job token that we acquired to start this job. Maybe we should track the job token in the job struct?

samu.c
```diff
@@ -90,6 +91,54 @@ jobsflag(const char *flag)
...
+	char *arg;
+
+	if (!env)
+		return;
+	env = xmemdup(env, strlen(env) + 1);
+
+	arg = strtok(env, " ");
+	/* first word might contain dry run */
+	if (arg && strchr(arg, 'n'))
+		buildopts.dryrun = true;
```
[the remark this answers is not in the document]
capezotte: That does seem to be what the first word is for, though the [protocol documentation](https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html) only goes as far as saying "check the first word for the letter `n`". Thanks for catching the oversight.

As for the second question, GNU Make (version 4.4.1 my system) invokes jobserver clients even in dry run mode.

samu.c
```diff
@@ -90,6 +91,54 @@ jobsflag(const char *flag)
...
+	int read_end, write_end;
+	struct pollfd check[2];
+
+	if (!flag)
+		return;
+	if (sscanf(flag, "%d,%d", &read_end, &write_end) == 2 && read_end >= 0 && write_end >= 0) {
+		/* assert valid fds */
+		check[0].fd = read_end;
+		check[1].fd = write_end;
+		if (poll(check, 2, 0) == -1 || check[0].revents & POLLNVAL || check[1].revents & POLLNVAL) {
```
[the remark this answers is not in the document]
capezotte: The documentation recommends doing the check. As it's there mostly to address a PEBKAC (user forgot a `+` in the Makefile), I do understand if you want to drop it.

build.c
capezotte: I think we can shuffle the tokens around. In fact, despite documentation saying "don’t assume that all tokens are the same character", in my testing they usually are.

build.c
```diff
@@ -545,20 +545,37 @@ void
...
+		tokenin.fd = buildopts.gmakepipe[0];
+	}
+
 	clock_gettime(CLOCK_MONOTONIC, &starttime);
 	formatstatus(NULL, 0);
 
 	nstarted = 0;
 	for (;;) {
+		/* see if GNU/make has authorized more jobs */
+		if (poll(&tokenin, 1, 0) > 0) {
```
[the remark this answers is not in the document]
capezotte: Polls are now unified (though fds is one larger than jobs now, which feels ugly).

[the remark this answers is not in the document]
michaelforney: Thanks!

samu.c
```diff
@@ -90,6 +91,50 @@ jobsflag(const char *flag)
...
+		return;
+	env = xmemdup(env, strlen(env) + 1);
+
+	arg = strtok(env, " ");
+	while (arg) {
+		if (arg[0] && arg[0] != '-' && strchr(arg, 'n')) {
+			buildopts.dryrun = true;
+		} else if (strncmp(arg, "-j", 2) == 0) {
+			jobsflag(arg + 2);
+		} else if (strncmp(arg, "--jobserver-auth=", 17) == 0 || strncmp(arg, "--jobserver-fds=", 16) == 0) {
```
michaelforney, line 130: I don't think we need to support `--jobserver-fds`. It's not documented, so I assume it's an old name for the flag.

samu.c
```diff
@@ -90,6 +91,50 @@ jobsflag(const char *flag)
...
+	int rfd, wfd;
+
+	if (!flag)
+		return;
+	if (sscanf(flag, "%d,%d", &rfd, &wfd) == 2) {
+		/* prepare error message */
+		errno = EBADF;
+	} else if (strncmp(flag, "fifo:", 5) == 0) {
+		rfd = open(flag + 5, O_RDONLY);
+		wfd = open(flag + 5, O_WRONLY);
```
michaelforney, line 106: You should just open it once as `O_RDWR` and set both `rfd` and `wfd` to the same value.

Also, we should print a message if this happens.

samu.c
```diff
@@ -90,6 +91,50 @@ jobsflag(const char *flag)
...
+static void
+jobserverflags(const char *flag)
+{
+	int rfd, wfd;
+
+	if (!flag)
+		return;
+	if (sscanf(flag, "%d,%d", &rfd, &wfd) == 2) {
+		/* prepare error message */
+		errno = EBADF;
```
michaelforney, line 103: This is a success case, right?

build.c
```diff
@@ -319,6 +327,15 @@ jobstart(struct job *j, struct edge *e)
 		warn("posix_spawn_file_actions_addclose:");
 		goto err3;
 	}
+	if (isjobserverclient()) {
+		/* do not allow children to steal GNU/tokens */
```
michaelforney, line 331: As @trws asked, do we need to prevent children from interacting with the job server?

If we do, I think it'd be better to set `FD_CLOEXEC` on the fds than closing them here.

build.c
```diff
@@ -522,43 +539,91 @@ jobwork(struct job *j)
 
 /* queries the system load average */
 static double
```
michaelforney, line 541: This should be `bool` now.

samu.c
capezotte: It's an old name for it (before GNU Make 4.2, I think). As far as I can tell, LTS distros don't use GNU Make 4.1.x anymore, so it can be safely dropped.

samu.c
capezotte: Print a message whenever a FIFO is used?

samu.c
capezotte: Yes. Since it's no longer validating the FDs with POLLNVAL below, setting errno here is just obsolete.

build.c
capezotte: I don't remember why I thought of that in the first place (nowhere in the documentation is this recommended).

After testing for a bit, it seems GNU Make uses the same `--jobserver-auth` parameters for every client, so it's fine for `samu` to share the jobserver with its own clients.

bonzini:
The main issue with this implementation is that if you have `make -j26` and your build is

```
      a     b     c                large files that take a huge time
       `----+----'
            d
   ,---.----+--------.
  e    f   g   ....   z            small files that take a small time
```

then samu will take all 26 tokens and block other users of the jobserver pipe. Instead, tokens must be taken only when jobs are started.

eli-schwartz:
@bonzini I haven't looked at the implementation at all. But, based on the previous review comment:
> One question I have, should a jobserver client request a token for every job it starts and release it when its done? Or is it fine if samurai takes as many jobs as it can use, and release them only when it has extra tokens that it can't use?

I had been under the assumption that in the case you describe, samu would relinquish 23 tokens as soon as it detects it is blocked on a/b/c and can't start d.

bonzini:
The only write system call that I see in the `jobserver` branch is

```
+                       /* must return the GNU/tokens once a job is done */
+                       if (buildopts.gmakepipe[1] > 0 && gmakelatest != gmaketokens) {
+                               if (write(buildopts.gmakepipe[1], --gmakelatest, 1) < 0) {
+                                       warn("write to jobserver:");
+                               }
+                               --maxjobs;
+                       }
```

so I don't think it does, though of course I might be wrong.

build.c
```diff
@@ -599,17 +656,30 @@ build(void)
...
 			if (jobs[i].failed)
 				++numfail;
+			/* return a token */
+			if (isjobserverclient() && gmakelatest > gmaketokens) {
+				if (write(buildopts.gmakepipe[1], --gmakelatest, 1) < 0)
+					warn("write to jobserver:");
+			}
+		}
+		if (fds[0].revents & POLLIN && isjobserverclient() && nstarted < ntotal && !reachedload()) {
+			gmakeread = read(buildopts.gmakepipe[0], gmakelatest, buildopts.maxjobs - numjobs);
```
bonzini, line 677: There is a race condition here:

* `poll` returns POLLIN
* another process steals the token that `poll` detected
* `read` blocks
* one or more children of samurai finish
* the whole build cannot reuse the token(s) currently owned by samurai

The solution used by GNU Make is complex and relies on [trapping SIGCHLD and dup-ing the file descriptor](https://make.mad-scientist.net/papers/jobserver-implementation/). It's complicated and it's why I just went for [reading the jobserver file descriptor in a separate thread](https://github.com/michaelforney/samurai/pull/104) instead.

capezotte: I'm not even sure GNU Make's solution is 100% robust without an extra `(void)alarm(1);`, or only starting a new process as an absolute last resort.

I really have to rethink this implementation.

capezotte:
I'm starting to think of an in-between approach of having the job spawn a helper program that waits for a token, feeds it back to samurai through a second pipe (so it's stored in the associated `job` struct, or the `gmakelatest` in this PR), and `exec`s the actual job. Only samurai would have read access to this second pipe, so there is no way of `POLLIN` getting outdated before `read`.

Once I have more time I might rework this PR into this design, but I'd like for feedback.

bonzini:
That seems similar to what I am doing, just with a process instead of a thread. You could also use fork() to do the wait, instead of using a helper program.

However, there would probably be more complications in returning the tokens on a call to `fatal()`. samurai does not attempt to kill child processes in that case, instead leaving them to die because of a broken pipe if they write anything to stdout/stderr; but in this case the children would also survive until they get a token. This would be untidy. And also, to ensure that the token is put back if the parent is dead, you would have to catch SIGPIPE (or temporarily ignore it so that you get EPIPE).

As an aside, it' s ok not to return the tokens on SIGINT/SIGTERM. I tried using a Makefile like

```
test-parent:
	$(MAKE) -f parent-makefile test-child1 test-child2 test-child3

test-child%:
```

Then invoking `make -j3 -Onone` and killing the child `make` in another terminal. Both SIGKILL and SIGTERM have the same behavior and show:

```
make[1]: *** [Makefile:10: test-child1] Terminated
make[1]: *** [Makefile:10: test-child3] Terminated
make[1]: *** [Makefile:10: test-child2] Terminated
make: *** [Makefile:7: test-parent] Terminated
make: INTERNAL: Exiting with 1 jobserver tokens available; should be 3!
```

So it's not a huge deal. But if you want to do it, a better way is to enhance samurai so that it traps SIGTERM and SIGINT, killing all the children in response. The basic idea would be as simple as

```
diff --git a/build.c b/build.c
index 65305d2..877d04c 100644
--- a/build.c
+++ b/build.c
@@ -494,6 +494,9 @@ jobwork(struct job *j)
 	size_t newcap;
 	ssize_t n;
 
+	if (killall)
+		goto kill;
+
 	if (j->buf.cap - j->buf.len < BUFSIZ / 2) {
 		newcap = j->buf.cap + BUFSIZ;
 		newdata = realloc(j->buf.data, newcap);
@@ -560,7 +563,7 @@ build(void)
 		if (buildopts.maxload)
 			maxjobs = queryload() > buildopts.maxload ? 1 : buildopts.maxjobs;
 		/* start ready edges */
-		while (work && numjobs < maxjobs && numfail < buildopts.maxfail) {
+		while (work && !killall && numjobs < maxjobs && numfail < buildopts.maxfail) {
 			e = work;
 			work = work->worknext;
 			if (e->rule != &phonyrule && buildopts.dryrun) {
@@ -616,10 +619,12 @@ build(void)
 		}
 		if (numjobs == 0)
 			break;
-		if (poll(fds, jobslen, 5000) < 0)
+		if (poll(fds, jobslen, 5000) < 0 && errno != EINTR)
 			fatal("poll:");
 		for (i = 0; i < jobslen; ++i) {
-			if (!fds[i].revents || jobwork(&jobs[i]))
+			if (!killall && !fds[i].revents)
+				continue;
+			if (jobwork(&jobs[i]))
 				continue;
 			--numjobs;
 			jobs[i].next = next;
```

where the killall variable is set by the signal handler. By terminating all jobs cleanly, the tokens would also be returned.

build.c
michaelforney: After reading that paper about the GNU make implementation, I'm convinced that we *must* use threads (or some sort of async I/O) to implement this correctly on samurai. The `SIGCHLD`/`EBADF` approach they use is clever, but won't work for us, since we buffer the output from jobs to avoid it getting jumbled.

The main issue is that you cannot predict whether a read on the jobserver fd will block or not. It doesn't matter for make since if it blocks, they have nothing to do anyway until a child exits, which can conveniently interrupt the read via `SIGCHLD`. samurai follows ninja in that it buffers job output, so to do the same trick we'd need the jobserver read to get interrupted when any of the pipes to jobs become readable.

Because of this, I'm going to close this PR in favor of #104. Thanks for your work getting started on this.

PR #104, opened 2024-04-19
Support for GNU Make jobserver (alternative implementation)

bonzini:
Alternative implementation to #94.

The main advantage is that the integration with the `build()` event loop is very clean, as it simply uses a pipe to signal the availability of tokens. Interacting with the job server is entirely embedded within a new `token.c` file that implements a simple API:

```
int tokeninit(void);
bool tokenget(struct edge *e);
void tokenput(void);
```

and on top of this, the integration is about 20 lines of code.

On the other hand `token.c` uses pthreads, which perhaps could be considered more complex. Waiting for reviews. :)


bonzini:
(rebased on top of #106, which provides "give back tokens on signals" behavior for free)

bonzini:
Sorry to be very explicit, but I would like a clear answer on whether the project is still maintained.

token.c
```diff
@@ -0,0 +1,334 @@
...
+ * retrieved */
+bool tokenget(struct edge *e)
+{
+	bool got_token = false;
+
+	if (parent_wfd == -1)
+		return true;
+
+	got_token = tokenread(local_rfd);
+	if (!got_token && !e->reserve) {
```
michaelforney, line 130: I'm worried about the following case:
1. samurai tries to run a job A, and does not get a token.
2. tokenget signals to the jobserver thread that it wants a token, and marks the reserve flag of A.
3. samurai enters poll
4. Some other job finishes and unblocks job B, adding it to the head of the `work` list.
5. The jobserver thread receives a token and writes it to the pipe
6. samurai tries to run job B, `tokenget` returns the token just written.
7. samurai tries to run job A. `tokenget` sees that the reserve flag is already set so does not request another job.
8. We don't start job A, even if there is another token available from the jobserver.

To solve this, it's almost sufficient to change this condition to `if (!e->reserve)`, but I think this doesn't work due to the free token. If you changed the jobserver wait condition to `while (pending_edges <= 1 && !done)` as well, that might do the trick.

Another option would be to keep track of the `work` list tail in build.c to make it operate FIFO.

bonzini: Yes, you're right! Making it FIFO is the simplest fix. Maybe another is adding a variable requested_tokens that is local to build.c.

I will take a look once I get round to testing and submitting the prerequisite SIGTERM patches.

RJVB: FWIW, it would be wonderful to have this! (Esp. if "official" Ninja does *not* already have jobserver support ;) )

PR #105, opened 2024-04-22
build samurai with Cosmopolitan Libc

ahgamut:
This PR adds a Github Action to build `samu` with Cosmopolitan Libc (cc @jart) when a new tag/release is created.

The result of the action is `samu.zip`, that contains `samu.com`, a fat binary that works on `x86_64`/`aarch64`, and debug ELF binaries. You can see it here: https://github.com/ahgamut/samurai/releases/tag/untagged-c654d227e531f9e3235f


eli-schwartz:
This seems like a rather large amount of effort to go to just for a non-portable binary that only runs on two of the most common Linux arches. Isn't the whole point of cosmopolitan libc to be, well, portable anywhere?

Would be more impressive if it actually lived up to its name.

jart:
@ahgamut Is it possible to build this project using just `cosmocc -O -o samurai *.c *.cc`?

ahgamut:
Ok, now it builds with just `cosmocc`, and the built binary `samu` should work on `x86_64` (Linux/BSDs/Windows/MacOS) and `aarch64` (Linux/MacOS).

PR #116, opened 2025-03-03
build: link with -lrt for clock_gettime

eli-schwartz:


michaelforney:
Thanks! I applied this a while ago but forgot to close.

michaelforney:
> The motivation for the change is that it fixed a build failure on Solaris.

Ah, ok, that makes sense.

> > what do you think about leaving $(LDLIBS) on the link command-line, but making it empty by default so that it works on most common unixes?
> 
> Well, that means on other systems it _fails_ by default, which is an unfortunate outcome for a tool that tries to stick to POSIX. So much for BSD being a POSIX operating system?

Checking again, it seems FreeBSD and NetBSD are fine, it's OpenBSD and macOS that had the problem.

Technically, OpenBSD doesn't even have the `c99`/`c17` wrappers, which is the part of POSIX that requires `-l rt` to work.

> I'm not sure what conclusion to draw from this. It sounds like a POSIX tool cannot compile successfully on BSD without including a configure script.
> 
> It may also be possible to write a Makefile link target that runs a more complicated shell script snippet that checks uname and skips -lrt on BAD systems known to not include it.

I'm a bit conflicted, too. I suspect more people are running samurai on OpenBSD and macOS than Solaris, so it would be nice if it worked by default with just `make`. I'm looking to make a release soon (yeah I know, it's about time), and am anticipating bug reports from people due to `-l rt`. However, Solaris isn't doing anything wrong, so it'd be a shame to break that in the process.

That said, since POSIX allows `clock_gettime()` to work without `-l rt`, which seems to be what most operating systems choose, it doesn't seem so bad to omit it by default, but make it easy to add it yourself with `make LDFLAGS=-lrt`. Right now though, I'm leaning towards leaving it as-is.

PR #118, opened 2025-05-14
Win32 support

cyanidle:
Hello. I am another poor soul trying to add windows support. 

I got basic stuff working: os-dependent jobs stuff is separated into os.h. I also fixed lots of memory leaks found with `-fsanitize-address`. Posix side with new abstractions seems to work OK.

Things that need fixing before windows support can be declared:
- [ ] `deps = msvc`
- [x] `$variable` substitutions are escaped with quotes, which works fine with `/usr/bin/sh -c`, but fails on Windows (CreateProcess does no CLI processing), when MSVC tries to compile something (`cl.exe ... -c '<file'`) -> causes `warning D9027 :  source file ''samurai\os-win32.c'' ignored`
- [ ] Parsing `.ninja_deps` from ninja causes segfault in `depsinit()` (node->gen is null for some reason)
- [ ] jobs/osjobs array seems to never shrink. On huge number of jobs in build this will cause lots of iterations over these two arrays. Perf hit is probably not very big, but this is very `O(n^2)`
- [ ] Makefile should be modified to support `.exe` suffix and use `mt.exe` to add manifest file -> important for UTF-8 and long paths. `NMake` does not support `Gnu Make`'s `ifeq()`, it has its own syntax for conditionals. May be easier to have two Makefiles, for `make` and `nmake`
- [x] `setvbuf()` is disabled in win32, bc it causes assert with bufsize zero
- [x] `canonpath()` and other path related APIS

CMakefile and related gitignores are for development only, it helps with IDE support.

There seem to be issues with formatting caused by said IDEs and TABS. Idk how to fix that :D

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,11 @@
 *.o
 .*.sw*
 /samu
+out/
+.vs/
+.vscode/
+.cache/
+build/
+CMakeUserPresets.json
+CMakePresets.json
+*.user
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,5 @@ out/
 build/
 CMakeUserPresets.json
 CMakePresets.json
+CMakeSettings.json
 *.user
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ set(SOURCE
 )
 
 if (WIN32)
-    list(APPEND SOURCE os-win32.c)
+    list(APPEND SOURCE os-win32.c samu.manifest)
 elseif(UNIX)
     list(APPEND SOURCE os-posix.c)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,29 @@
+cmake_minimum_required(VERSION 3.16)
+project(samurai C)
+
+set(SOURCE
+    build.c
+    deps.c
+    env.c
+    graph.c
+    htab.c
+    log.c
+    parse.c
+    samu.c
+    scan.c
+    tool.c
+    tree.c
+    util.c
+)
+
+if (WIN32)
+    list(APPEND SOURCE os-win32.c)
+elseif(UNIX)
+    list(APPEND SOURCE os-posix.c)
+endif()
+
+add_executable(samu ${SOURCE})
+
+if (UNIX)
+    target_link_libraries(samu PRIVATE rt)
+endif()
--- a/build.c
+++ b/build.c
@@ -114,8 +114,7 @@ buildadd(struct node *n)
 {
 	struct edge *e;
 	struct node *newest;
-	size_t i;
-	bool generator, restat;
+    size_t i;
 
 	e = n->gen;
 	if (!e) {
@@ -153,15 +152,19 @@ buildadd(struct node *n)
 			++e->nblock;
 	}
 	/* all outputs are dirty if any are older than the newest input */
-	generator = edgevar(e, "generator", true);
-	restat = edgevar(e, "restat", true);
+    struct string* generator = edgevar(e, "generator", true);
+    struct string* restat = edgevar(e, "restat", true);
 	for (i = 0; i < e->nout && !(e->flags & FLAG_DIRTY_OUT); ++i) {
 		n = e->out[i];
-		if (isdirty(n, newest, generator, restat)) {
+        if (isdirty(n, newest, (bool)generator, (bool)restat)) {
 			n->dirty = true;
 			e->flags |= FLAG_DIRTY_OUT;
 		}
 	}
+    if (generator)
+        free(generator);
+    if (restat)
+        free(restat);
 	if (e->flags & FLAG_DIRTY) {
 		for (i = 0; i < e->nout; ++i) {
 			n = e->out[i];
@@ -265,6 +268,7 @@ printstatus(struct edge *e, struct string *cmd)
 	formatstatus(status, sizeof(status));
 	fputs(status, stdout);
 	puts(description->s);
+    free(description);
 }
 
 static int
@@ -289,6 +293,8 @@ jobstart(struct osjob_ctx *osctx, struct osjob *oj, struct job *j, struct edge *
 			goto err0;
 	}
 	j->edge = e;
+    if (j->cmd)
+        free(j->cmd);
 	j->cmd = edgevar(e, "command", true);
 
 	if (!consoleused)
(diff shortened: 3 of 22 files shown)


os-win32.c
```diff
@@ -0,0 +1,333 @@
...
+		warn("job failed with status %d: %s", exit_code, cmd->s);
+		goto err;
+	}
+	return osjob_close(osctx, ojob);
+err:
+	osjob_close(osctx, ojob);
+	return -1;
+}
+
+int osjob_close(struct osjob_ctx* osctx, struct osjob* ojob)
```
TomasBorquez, line 325: maybe `__attribute__((unused))` on `osctx` or `_osctx`

os-win32.c
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	return 0;
+}
+
+int osjob_wait(struct osjob_ctx* osctx, struct osjob ojobs[], size_t jobs_count, int timeout)
+{
+	OVERLAPPED_ENTRY entries[64];
+	ULONG num_entries = 0;
+	const DWORD timeout_ms = timeout == -1 ? INFINITE : timeout;
```
TomasBorquez, line 260: type aliasing:
```c
const DWORD timeout_ms = timeout == -1 ? INFINITE : (DWORD)timeout;
```

TomasBorquez (commented):
left comments on some warning you were getting, also i think you can fix the indentation with clang-format project settings:
```bash
clang-format -i -style=file $(find ./ -name "*.c" -o -name "*.h")
```

os-win32.c
```diff
@@ -0,0 +1,333 @@
...
+
+	return 0;
+}
+
+///////////////////////////// JOBS
+
+static HANDLE 
+win_create_nul()
+{
+	SECURITY_ATTRIBUTES sa = {sizeof(sa)};
```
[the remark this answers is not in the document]
cyanidle: How is this warning called? I dont think it should be used. sizeof(struct) is always the first field. It is actually really nice to be able to do `T config = {sizeof(config};`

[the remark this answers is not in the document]
TomasBorquez: you are right, the warning is `-Wmissing-field-initializers`, you can create a .clangd or add it to the makefile as:

`-Wno-missing-field-initializers`

TomasBorquez:
## First test - simplest
Simple ninja file:

```ninja
cc = gcc
target = main.exe

rule link
  command = $cc $flags -o $out $in

rule compile
  command = $cc $flags -c $in -o $out

build main.o: compile main.c
build $target: link main.o

default $target
```

Ran it on the same folder:
```bash
Lewboski@DESKTOP-CK46RK4 MINGW64  /c/Users/Lewboski/Desktop/Programming/learn/mate/tests/01-basic-build/build
$ ./samu.exe 
[1/2] gcc  -c main.c -o main.o
C:\Users\Lewboski\Desktop\Programming\learn\mate\tests\01-basic-build\build\samu.exe: CreateProcess: No such file or directory
C:\Users\Lewboski\Desktop\Programming\learn\mate\tests\01-basic-build\build\samu.exe: job failed to start
```

## Second test - vars
I tested it with some variables on a single source:
```ninja
cc = gcc
flags = -Wall -g
cwd = C$:\Users\Lewboski\Desktop\Programming\learn\mate\tests\01-basic-build
builddir = C$:\Users\Lewboski\Desktop\Programming\learn\mate\tests\01-basic-build\build
target = $builddir/main.exe

rule link
  command = $cc $flags -o $out $in

rule compile
  command = $cc $flags -c $in -o $out

build $builddir/main.o: compile $cwd/src/main.c
build $target: link $builddir/main.o

default $target
```

Ran it with -f, and got:
```bash
Lewboski@DESKTOP-CK46RK4 MINGW64  /c/Users/Lewboski/Desktop/Programming/learn/mate/tests/01-basic-build
$ ./build/samu.exe -f ./build/build.ninja
C:\Users\Lewboski\Desktop\Programming\learn\mate\tests\01-basic-build\build\samu.exe: mkdirs C:\: No error
C:\Users\Lewboski\Desktop\Programming\learn\mate\tests\01-basic-build\build\samu.exe: job failed to start
C:\Users\Lewboski\Desktop\Programming\learn\mate\tests\01-basic-build\build\samu.exe: subcommand failed
```

you can use these as reference meanwhile testing

.gitignore
```diff
@@ -1,3 +1,11 @@
...
 .*.sw*
 /samu
+out/
+.vs/
+.vscode/
+.cache/
+build/
+CMakeUserPresets.json
+CMakePresets.json
+*.user
```
michaelforney, line 11: These are probably better to put in your `.git/info/exclude` instead.

build.c
```diff
@@ -23,17 +18,15 @@ struct job {
...
+        size_t next;
 	bool failed;
 };
 
 struct buildoptions buildopts = {.maxfail = 1};
 static struct edge *work;
 static size_t nstarted, nfinished, ntotal;
 static bool consoleused;
-static struct timespec starttime;
+static struct ostimespec starttime;
```
michaelforney, line 29: I think we should just make it `ostime()` and have it return a `double`.

CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+cmake_minimum_required(VERSION 3.16)
```
michaelforney, line 1: Do you think it's possible to get the Makefile to work with `nmake`?

build.c
```diff
@@ -362,18 +368,19 @@ edgedone(struct edge *e)
 {
 	struct node *n;
 	size_t i;
-	struct string *rspfile;
-	bool restat;
+    struct string *rspfile;
 	int64_t old;
 
-	restat = edgevar(e, "restat", true);
+    struct string * restat = edgevar(e, "restat", true);
```
michaelforney, line 374: Can you show the error here? Does MSVC not allow conversion of pointer to bool?

michaelforney (commented):
This looks like a pretty good initial attempt! I'll need to take a closer look at the jobs abstraction.

It seems like there are a lot of places where pointer to bool conversion doesn't work. Is this because MSVC doesn't have bool? Or is something else going on?

Could you revert the formatting changes? As mentioned in `.clang-format`, the changes it suggests aren't authoritative. I have formatted everything how I want it, and I think the rules in `.clang-format` are as close as I can get.

There should be a way to configure your IDE to use tabs. Maybe somewhere in the settings?

CMakeLists.txt
cyanidle: NMake is not very compatible with Make, easier to maintain both, considering project simplicity

.gitignore
cyanidle: Agree, will revert later probably

build.c
cyanidle: MSVC allows, I did this and similar patches to fix memory leaks - edgevar() allocates new string (e.g: see line 380).

cyanidle:
@michaelforney Hello! you are finally active. Though I myself moved away from this PR) I kinda lost interest and there is a lot of work to be done.
But maybe not all hope is lost, if you think this job abstraction thing is worth it - I may continue work on Win compatability. Though not at a very quick pace.

cyanidle:
I read couple of your new commits, I see you are doing some work on abstracting os-specific things. Reviving this PR may not be very worth it (though looking at the implementation of some things probably is)

Issue #20, opened 2019-04-29
Issues with logging samu with tee(1)


orbea:
OS: `Slackware64-14.2`
samu: `0.6`
ninja: `1.9.0`
meson: `0.49.0`
dxvk: `1.0.3`

I am trying to build `dxvk` using `meson` with `samu` instead of `ninja` on Slackware 14.2 which results in a compile failure with both `samu` and `ninja`.

I am logging the output of the build script by piping it through `tee(1)`, but there are two potential problems with `samu` which `ninja` does not have. I have attached two build logs which I hope illustrate the problem. In short the log created with `ninja` is short and easy to read while the log with `samu` is giant, has broken new lines and is very hard to read. The end result is that I must use `ninja` to report the issue to `dxvk`.

The exact method of logging is.
```
./foo.SlackBuild 2>&1 | tee foo.log
```
Where `foo.SlackBuild` is a shell script which wraps `meson` and `ninja` or `samu`.

[dxvk.ninja.log](https://github.com/michaelforney/samurai/files/3128560/dxvk.ninja.log)
[dxvk.samu.log](https://github.com/michaelforney/samurai/files/3128561/dxvk.samu.log)

michaelforney:
The two error error messages are different because they are from different source files. samurai does not guarantee that it will execute the build in the same order as ninja, just that it will be a valid order according to the build manifest. I suspect that if you explicitly built the `src/dxbc/5c53ffa@@dxbc@sta/dxbc_util.cpp.o` target with ninja, you'd see the same error (or `src/util/ed6d25d@@util@sta/util_env.cpp.o` with samu).

I don't see the broken newlines you are talking about. Perhaps you are referring to the embedded ANSI escape codes? This is because the CFLAGS its using include `-fdiagnostics-color=always` which means that gcc is printing the escape codes even though the output is not a terminal. Perhaps ninja is doing something extra to strip them?

Yep, that appears to be what ninja is doing: https://github.com/ninja-build/ninja/blob/master/src/build.cc#L147-L162

Looks like meson is the one adding the `-fdiagnostics-color=always` flag, and you can disable it with `meson -Db_colorout=auto`.

orbea:
Thanks for the explanation about the build errors, that helps me understand much better!

> Perhaps you are referring to the embedded ANSI escape codes? 

[...]

PR #120, opened 2026-04-01
Strip ANSI escape codes when not writing to a terminal

fredrikfoss:
ANSI color escape codes are stripped from job output when standard output is redirected to a pipe or file, unless CLICOLOR_FORCE is set to a value other than "0", or TERM is set to "dumb". This matches ninja's behavior.

See <https://github.com/michaelforney/samurai/issues/20#issuecomment-4156981975>. I'll just open the pull request before I forget.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,12 +39,6 @@ are several cases where it is slightly different:
   so the first to execute depends on the address returned by `malloc`.
   This may result in build failures due to insufficiently specified
   dependencies in the project's build system.
-- samurai does not post-process the job output in any way, so if it
-  includes escape sequences they will be preserved, while ninja strips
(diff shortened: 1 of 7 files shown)


fredrikfoss:
> That said, I think your implementation of escape sequence stripping is pretty good. I'm open to putting it in a branch if people want this behavior.

Thanks, but in that case I think I would personally just stick to using the default version. I'm assuming it's the one package managers will ship.

By the way, I was pretty conservative in the implementation as it is based on ninja's. It can be hardened quite a bit more. For example, right now it ignores OSC sequences so hyperlinks won't be stripped (although I'm not sure any compilers emit those?) It shouldn't require much change, I have a working draft.

PR #121, opened 2026-04-06
Only link librt on Linux

chenrui333:
`samurai` 1.3 links with `-lrt` unconditionally, which breaks macOS because Darwin does not provide `librt`.

This keeps `-lrt` on Linux and skips it on other platforms.

- https://github.com/Homebrew/homebrew-core/pull/276193

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ PREFIX=/usr/local
 BINDIR=$(PREFIX)/bin
 MANDIR=$(PREFIX)/share/man
 ALL_CFLAGS=$(CFLAGS) -std=c99 -Wall -Wextra -Wshadow -Wmissing-prototypes -Wpedantic -Wno-unused-parameter
-LDLIBS=-lrt
+LDLIBS=
 OBJ=\
 	build.o\
 	deps.o\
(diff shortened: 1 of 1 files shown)


eli-schwartz:
This patch is 100% incorrect (this is not being linked for Linux support) -- please see the discussion on the PR that added librt:


_Originally posted by @eli-schwartz in https://github.com/michaelforney/samurai/issues/116#issuecomment-4072326018_
            
> The motivation for the change is that it fixed a build failure on Solaris.
> 
> > what do you think about leaving $(LDLIBS) on the link command-line, but making it empty by default so that it works on most common unixes?
> 
> Well, that means on other systems it *fails* by default, which is an unfortunate outcome for a tool that tries to stick to POSIX. So much for BSD being a POSIX operating system?
> 
> I'm not sure what conclusion to draw from this. It sounds like a POSIX tool cannot compile successfully on BSD without including a configure script.
> 
> It may also be possible to write a Makefile link target that runs a more complicated shell script snippet that checks uname and skips -lrt on BAD systems known to not include it.

chenrui333:
I am open to new ideas, but it indeed fails on macos?

eli-schwartz:
> I am open to new ideas, but it indeed fails on macos?

Yup -- I don't think anyone argued against that. Just that the fix as proposed goes way too far in the other direction.

macOS is likely displaying the issue as inherited from BSD, which was also noted to have this issue. My personal idea would be to have the Makefile target check if -lrt successfully links and remove -lrt and try again if it failed. This should theoretically be robust against any theoretical platform that displays the same issue but with further uname identities.

michaelforney:
I added a note about this with the suggested fix in the README. Hopefully this helps anyone else who encounters this issue.

